PR #468, opened 2023-08-28
chore: prepare for new release.

xsahil03x:


deven98 (changes requested):
Can we also update versions for dogfooding and examples?

xsahil03x:
> Can we also update versions for dogfooding and examples?

@deven98 we don't have them published anywhere, so not really needed.

deven98:
> > Can we also update versions for dogfooding and examples?
> 
> @deven98 we don't have them published anywhere, so not really needed.

Let's change them regardless, for the sake of parity and testing.

deven98 (approved):
LGTM

PR #481, opened 2023-09-16
Hide debug stats in production

Nash0x7E2:
This PR tweaks the debug stats visibility to ensure stats are only shown when the app is in debug mode.


kanat (approved):
LGTM

PR #488, opened 2023-09-20
FVS-84: align SDK initialization

kanat:
### 🎯 Goal

Closes: https://stream-io.atlassian.net/browse/FVS-84

#### General

### ☑️Reviewer Checklist
- [ ] Sample runs & works


packages/stream_video/lib/src/coordinator/open_api/coordinator_client_open_api.dart
```diff
@@ -76,26 +84,75 @@ class CoordinatorClientOpenApi extends CoordinatorClient {
...
+      timeLimit: const Duration(seconds: 15),
+    )
+        .then((it) {
+      if (it.isConnected) {
+        _logger.v(() => '[waitUntilConnected] completed: $it');
+        return const Result.success(none);
+      } else {
+        _logger.e(() => '[waitUntilConnected] failed: $it');
+        return Result<None>.error('waitUntilConnected failed');
+      }
```
xsahil03x, line 149: We are only listening for `it.isConnected` then will there be any case it can be false?

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -84,39 +129,16 @@ class StreamVideo {
...
   /// API key.
-  static Future<void> reset({bool disconnectUser = false}) async {
-    if (disconnectUser) {
-      await _instanceHolder.instance.disconnectUser();
+  static Future<void> reset({bool disconnect = false}) async {
+    if (disconnect) {
+      await _instanceHolder.instance.disconnect();
     }
     return _instanceHolder.reset();
   }
```
xsahil03x, line 144: `_instanceHolder` won't always contain the instance as we now have it configurable.
```dart
options.installAsSingleton
```

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -173,46 +195,72 @@ class StreamVideo {
...
+        await pushNotificationManager?.onUserLoggedIn();
+      } catch (e, stk) {
+        _logger.w(() => '[connect] #pnManager; failed: $e, $stk');
+      }
+      try {
+        _logger.v(() => '[connect] checking token');
+        await _tokenManager.getToken();
+      } catch (e, stk) {
+        _logger.w(() => '[connect] #tokenManager; failed: $e, $stk');
+      }
```
xsahil03x, line 225: any reason we are fetching token here? The `connectUser` call will do it automatically right?

kanat: We just want to wait if it's not fetched yet to let `connect` complete when token is available

kanat: GetToken does not trigger extra requests.
It either triggers request if token is null or waits the previous request to complete.
If tokens exists it will return immediately

packages/stream_video/lib/src/coordinator/open_api/coordinator_client_open_api.dart
kanat: Good catch

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -173,46 +189,83 @@ class StreamVideo {
...
+        await pushNotificationManager?.onUserLoggedIn();
+      } catch (e, stk) {
+        _logger.w(() => '[connect] #pnManager; failed: $e, $stk');
+      }
+      return const Result.success(none);
+    } catch (e, stk) {
+      _logger.e(() => '[connect] failed(${user.id}): $e');
+      return Result.failure(VideoErrors.compose(e, stk));
+    }
+  }
```
xsahil03x, line 226: I think the `connect` call should store the future and return with the previous one if already called. Otherwise calling connect will add subscriptions again and will also register device on every call.

PR #500, opened 2023-10-06
Anonymous users should not connect to the WS

kanat:
### 🎯 Goal

Align with other SDKs

#### General

### ☑️Reviewer Checklist
- [ ] Sample runs & works

### DIFF
--- a/packages/stream_video/lib/src/stream_video.dart
+++ b/packages/stream_video/lib/src/stream_video.dart
@@ -253,7 +253,9 @@ class StreamVideo {
   Future<Result<UserToken>> connect() async {
     if (currentUserType == UserType.anonymous) {
       _logger.w(() => '[connect] rejected (anonymous user)');
-      return Result.error('Cannot connect anonymous user to the WS');
+      return Result.error(
+        'Cannot connect anonymous user to the WS due to Missing Permissions',
+      );
     }
     _connectOperation ??= _connect().asCancelable();
     return _connectOperation!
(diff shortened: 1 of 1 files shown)


packages/stream_video/lib/src/stream_video.dart
```diff
@@ -251,6 +251,10 @@ class StreamVideo {
 
   /// Connects the user to the Stream Video service.
   Future<Result<UserToken>> connect() async {
+    if (currentUserType == UserType.anonymous) {
+      _logger.w(() => '[connect] rejected (anonymous user)');
+      return Result.error('Cannot connect anonymous user to the WS');
```
Nash0x7E2, line 256: Should this be considered an erorr, @kanat?

For example, if I am following the tutorial and decide to connect as a guest, an error here might be a deceiving.

kanat: There are 3 user types:
- authenticated
- guest
- anonymous

For `anonymous` we should not connect to the WS.

This logic is common for all SDKs.
Here is the iOS [implementation](https://github.com/GetStream/stream-video-swift/blob/main/Sources/StreamVideo/StreamVideo.swift#L503-L507)

Nash0x7E2: That's fair. Can we update the text there to be a bit more descriptive? 

iOS for example uses "Missing Permissions" as the error message.

kanat: Updated 👍

Nash0x7E2 (approved):
LGTM

PR #503, opened 2023-10-18
fix: call controls in landscape

Brazol:
Fixes missing call controls in landscape mode

### DIFF
--- a/packages/stream_video_flutter/lib/src/call_controls/call_controls.dart
+++ b/packages/stream_video_flutter/lib/src/call_controls/call_controls.dart
@@ -63,20 +63,13 @@ class StreamCallControls extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    if (isMobileLandscape(context)) {
-      return LandscapeCallControls(
-        options: options,
-        spacing: spacing,
-      );
-    } else {
-      return PortraitCallControls(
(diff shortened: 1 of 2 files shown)


packages/stream_video_flutter/lib/src/call_controls/call_controls.dart
```diff
@@ -63,20 +63,13 @@ class StreamCallControls extends StatelessWidget {
...
-        borderRadius: borderRadius,
-      );
-    }
+    return PortraitCallControls(
+      options: options,
+      elevation: elevation,
+      spacing: spacing,
+      padding: padding,
+      borderRadius: borderRadius,
+    );
```
kanat, line 72: Just curious, we no longer support both landscapes? :)

Brazol: The landscape support for call controls was introduced in [this PR](https://github.com/GetStream/stream-video-flutter/pull/257) but then most of it was removed in [this one](https://github.com/GetStream/stream-video-flutter/pull/338) for some reason. The effect was that in landscape no controls were shown. 
This PR just cleans the rest of the logic so at least controls are shown. If we want to support landscape mode it would require a separate task to reimplement the removed logic.

PR #504, opened 2023-10-19
fix: end CallKit call on call ending

Brazol:


esarbanis:
LTGM, but would like @deven98 to sign it off as well :sweat_smile:

kanat (approved):
LGTM

PR #508, opened 2023-10-26
fix: fixes cast issue for `capabilities_by_role` in `CoordinatorCallUpdatedEvent`

deven98:
Previously, any update event would lead to a cast issue leading to `capabilities_by_role` not being cast correctly. This PR fixes that issue and update events no longer cause an error.


kanat (approved):
LGTM

PR #513, opened 2023-11-06
feat: Added customization of incoming/outgoing call screen

Brazol:
Added builders for incoming/outgoing call screens to customize participants and avatars.
Added incoming display name override for CallKit
Fixed missing app icon in CallKit call screen


deven98 (approved):
LGTM

PR #533, opened 2023-11-20
fix: `connectUser` / auto connect bug 

deven98:
Adds a `includeUserDetails` field to determine if user should be passed to backend.

By default, both `connectUser` and auto-connecting default to passing fields to backend.


Brazol (approved):
LGTM

PR #535, opened 2023-11-21
fix: doc updates

deven98:
Fixes existing docs


Nash0x7E2:
@deven98 - I am going to assume this PR is now stale, shall we close it?

PR #540, opened 2023-12-01
docs: added additional cookbook guides

deven98:
Added additional cookbook guides:

* Call reactions
* Video fallback
* Call lobby
* Incoming/outgoing


Nash0x7E2:
@deven98 - what is the status of this PR? If we are not planning to update it, let's close it.

PR #544, opened 2023-12-06
feat: ringing call cancelation

Brazol:
Added automatic call cancelation of ringing call when caller is ending the call.

### DIFF
--- a/packages/stream_video_push_notification/lib/src/stream_video_push_notification.dart
+++ b/packages/stream_video_push_notification/lib/src/stream_video_push_notification.dart
@@ -12,6 +12,10 @@ import 'stream_video_push_params.dart';
 
 part 'stream_video_push_provider.dart';
 
+const _idToken = 1;
+const _idCallIncoming = 2;
+const _idCallEnded = 3;
+
 /// Implementation of [PushNotificationManager] for Stream Video.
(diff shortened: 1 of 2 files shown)


packages/stream_video_push_notification/lib/src/stream_video_push_notification.dart
```diff
@@ -45,7 +50,33 @@ class StreamVideoPushNotificationManager implements PushNotificationManager {
...
+    _subscriptions.add(
+      _idCallAcceptDecline,
+      onCallEvent.whereType<ActionCallAccept>().map((_) => null).mergeWith(
+          [onCallEvent.whereType<ActionCallDecline>().map((_) => null)]).listen(
+        (_) {
+          _subscriptions.cancel(_idCallEnded);
+        },
+      ),
+    );
+  }
```
kanat, line 79: Hey @Brazol,

I was wondering if we can start listening for these events inside `registerDevice` function 🙂

Brazol: We have to listen to them in the constructor for it to work when the app is closed

deven98 (approved):
LGTM

PR #554, opened 2023-12-14
fix: push notification cancelation

Brazol:


packages/stream_video/example/lib/main.dart
```diff
@@ -0,0 +1,11 @@
...
+import 'package:stream_video/stream_video.dart';
+
+Future<void> main() async {
+  WidgetsFlutterBinding.ensureInitialized();
+
+  StreamVideo.create(
+    'api_token',
+    user: User.guest(userId: 'user_id'),
+  );
+}
```
deven98, line 11: Is this an example for pub?

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -572,13 +572,18 @@ class StreamVideo {
     final callCid = payload['call_cid'] as String?;
     if (callCid == null) return false;
 
-    final uuid = const Uuid().v4();
+    var callId = const Uuid().v4();
+    final splitCid = callCid.split(':');
+    if (splitCid.length == 2) {
+      callId = splitCid.last;
+    }
+
```
deven98, line 580: Are we sure it's in the correct format? I remember this giving me some troubles earlier.

packages/stream_video_push_notification/lib/src/stream_video_push_notification.dart
```diff
@@ -54,14 +55,26 @@ class StreamVideoPushNotificationManager implements PushNotificationManager {
...
+            ),
+          );
+
+          _subscriptions.add(
+            _idCallAccepted,
+            client.events.on<CoordinatorCallAcceptedEvent>(
+              (event) async {
+                await FlutterCallkitIncoming.silenceEvents();
+                await FlutterCallkitIncoming.endCall(event.callCid.id);
+                await Future<void>.delayed(const Duration(milliseconds: 300));
```
deven98, line 76: Why do we need to delay here?

packages/stream_video/example/lib/main.dart
Brazol: yes, we had it before just like this (probably just for the pub scoring) then it got removed and scoring with it. So now it's back :)

packages/stream_video/lib/src/stream_video.dart
Brazol: yes, the issue is with iOS CallKit requiring it to be valid uuid, which it is when we split it and use the id part and not the whole cid

packages/stream_video_push_notification/lib/src/stream_video_push_notification.dart
Brazol: To account for asynchronous race, it happened that events were unsilenced before the endCall event was triggered on the native side.

packages/stream_video/example/lib/main.dart
deven98: This is fine for now but let's add these things in separate PRs so we don't confuse stuff.

packages/stream_video/lib/src/stream_video.dart
deven98: For some reason, in the early video time it sometimes wasn't a valid UUID, good to know it is now. This is great, thanks.

packages/stream_video_push_notification/lib/src/stream_video_push_notification.dart
deven98: Fair enough.

deven98 (approved):
LGTM

PR #560, opened 2023-12-21
add custom data to call participants

Brazol:
Adds missing custom field with custom user data to CallParticipantState.


kanat (approved):
LGTM

PR #561, opened 2023-12-21
default background for call screen on image error

Brazol:
Shows default background image for incoming/outgoing call screens on `participant.image` error.


deven98 (approved):
LGTM

PR #570, opened 2024-01-18
feat: added background voip handle called in terminated app state when voip push is received

Brazol:
### 🎯 Goal

When iOS app is inactive and receives voip push it will now start Flutter engine to connect to StreamVideo and listen to incoming events. This will fix ringing not canceling when caller hangs up issue.


docusaurus/docs/Flutter/05-advanced/02-ringing.mdx
```diff
@@ -378,7 +378,49 @@ say you want to end all calls on the CallKit side, you can end them this way:
 StreamVideo.instance.pushNotificationManager?.endAllCalls();
 ```
 
-#### Step 6 - Add native code to the iOS project
+#### Step 6 - Add callback to handle call in terminated state
+
+When the iOS app is terminated the Flutter engine is not running and we need to start it up to handle the Stream call events. This is done internally by the SDK, the only thing needed (similarly to Android) is to register a callback handle that will run when Flutter engine starts and connect to StreamVideo.
```
deven98, line 383:
```suggestion
When an iOS app is terminated, the Flutter engine is not running. We need to start up the engine to handle Stream call events whenever a call is received by the app. The Stream SDK performs the job of running a Flutter engine instance whenever a call is received. However, on the app side, you need to register a callback handle that will connect to `StreamVideo`. 
```

dogfooding/lib/screens/home_screen.dart
```diff
@@ -75,6 +78,56 @@ class _HomeScreenState extends State<HomeScreen> {
     }
   }
 
+  Future<void> _directCall(BuildContext context) async {
```
deven98, line 81: Do we want to add dogfooding changes in the same PR?

packages/stream_video_push_notification/ios/Classes/StreamVideoPKDelegateManager.swift
```diff
@@ -8,7 +8,7 @@ public class StreamVideoPKDelegateManager: NSObject, PKPushRegistryDelegate {
     
     private var pushRegistry: PKPushRegistry?
     private var defaultData: [String: Any]?
-    private var methodChannel: FlutterMethodChannel?
+    private var mainChannel: FlutterMethodChannel?
```
deven98, line 11: What was the reason behind the name change?

packages/stream_video_push_notification/ios/Classes/StreamVideoPKDelegateManager.swift
```diff
@@ -44,13 +44,35 @@ public class StreamVideoPKDelegateManager: NSObject, PKPushRegistryDelegate {
             return completion()
         }
         
+        print("voip received")
```
deven98, line 47: Should we remove the prints or are they there for debugging?

packages/stream_video_push_notification/ios/Classes/StreamVideoPKDelegateManager.swift
Brazol: It was from when I tried to create separate `backgroundChannel`. If it's not a big problem I would leave the `mainChannel` name since we might need the `backgroundChannel` anyway in the future.

dogfooding/lib/screens/home_screen.dart
Brazol: hmm good question. On one side it would be nice to have a core SKD change as a separate PR on the other side we can't test it without the dogfooding changes ;)

deven98: Let's merge it for now mainly because push is notoriously difficult to test on for now. Keep in mind someone might ask us to remove this in the future to adhere to the design and we may need to comply.

deven98: Also, this is a unique example but in the future let's merge these separately since we can revert specific commits much more easily.

PR #579, opened 2024-02-01
fix: fixed rtc tracks when switching between screen share types on ios

Brazol:


packages/stream_video/lib/src/webrtc/rtc_manager.dart
```diff
@@ -714,6 +714,21 @@ extension RtcManagerTrackHelper on RtcManager {
 
     // Track found, mute/unmute it.
     if (track != null) {
+      if (enabled &&
+          track.trackType == SfuTrackType.screenShare &&
```
esarbanis, line 718: Can you somehow encapsulate this long condition behind a getter?

Brazol: Done

esarbanis (approved):
LGTM

PR #580, opened 2024-02-07
feat: push notification enhancements

Brazol:
### 🎯 Goal

- Fixes late push notification handling on Android, where already ended call was ringing if the device was offline and the push was delivered with a delay
- Call is also now ending correctly when caller times out while calling

### DIFF
--- a/packages/stream_video/lib/src/stream_video.dart
+++ b/packages/stream_video/lib/src/stream_video.dart
@@ -639,30 +639,28 @@ class StreamVideo {
           return CallRingingState.ended;
         }
 
-        if (_state.currentUser.name != null) {
-          if (callData.metadata.session.acceptedBy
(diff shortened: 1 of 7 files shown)


packages/stream_video/lib/src/stream_video.dart
```diff
@@ -574,24 +576,98 @@ class StreamVideo {
...
+          }
+
+          if (callData.metadata.session.rejectedBy
+              .containsKey(_state.currentUser.name)) {
+            _logger.e(() => '[getCallRingingState] call already rejected');
+            return CallRingingState.rejected;
+          }
+
+          final otherMembers = callData.metadata.members.keys.toList()
+            ..remove(_state.currentUser.name);
```
deven98, line 656: Check if this is supposed to be name or id.

deven98 (approved):
LGTM

PR #583, opened 2024-02-15
feat: Refreshed call screen of dogfooding app 

Brazol:
Part 2 of Dogfooding app refresh merge after https://github.com/GetStream/stream-video-flutter/pull/582

PR #584, opened 2024-02-16
feat: Stream video flutter UI refresh - removed participant info components

Brazol:
This is part 3 of the refresh - merge after https://github.com/GetStream/stream-video-flutter/pull/583


dogfooding/lib/screens/call_participants_list.dart
```diff
@@ -0,0 +1,129 @@
...
+}
+
+class _CallParticipantsListState extends State<CallParticipantsList> {
+  StreamSubscription<CallState>? _callStateSubscription;
+
+  @override
+  void initState() {
+    super.initState();
+
+    _callStateSubscription = widget.call.state.listen(_setState);
```
esarbanis, line 26: Why don't we wrap this `call.state` in a builder widget and drop this `StatefulWidget` to `StatelessWidget`?

dogfooding/lib/screens/lobby_screen.dart
```diff
@@ -14,17 +19,138 @@ class LobbyScreen extends StatelessWidget {
   final ValueChanged<CallConnectOptions> onJoinCallPressed;
   final Call call;
 
+  @override
+  State<LobbyScreen> createState() => _LobbyScreenState();
+}
+
+class _LobbyScreenState extends State<LobbyScreen> {
+  RtcLocalAudioTrack? _microphoneTrack;
+  RtcLocalCameraTrack? _cameraTrack;
```
esarbanis, line 28: I don't understand why we need to keep track of these variables in this Widget. Can you please elaborate a bit?

dogfooding/lib/widgets/call_duration_title.dart
```diff
@@ -0,0 +1,82 @@
...
+      _duration = DateTime.now().difference(startedAt);
+
+      _timer = Timer.periodic(const Duration(seconds: 1), (timer) {
+        if (!mounted) {
+          timer.cancel();
+          return;
+        }
+
+        setState(() {
+          _duration += const Duration(seconds: 1);
```
esarbanis, line 42: I think it is more accurate to keep `startedAt` in the state and not the `_duration` and then calculate the `difference`. This way we can provision for any unsynced ticks.

dogfooding/lib/widgets/stream_button.dart
esarbanis, line 1: Nice work here!

packages/stream_video_flutter/lib/src/call_controls/controls/toggle_layout_option.dart
```diff
@@ -0,0 +1,52 @@
+import 'package:flutter/material.dart';
+
+import '../../../stream_video_flutter.dart';
+import '../../theme/stream_icons.dart';
```
esarbanis, line 4: I think we should avoid relative imports and either import from package or rollup files.

dogfooding/lib/screens/call_participants_list.dart
Brazol: Good idea, done

dogfooding/lib/screens/lobby_screen.dart
Brazol: Those are created and destroyed by `StreamLobbyVideo` when turning camera/microphone on/off and have to be stored so that later when call is created they are passed in CallConnectoptions to initialize call screen

dogfooding/lib/widgets/call_duration_title.dart
Brazol: yes, agreed. Done

packages/stream_video_flutter/lib/src/call_controls/controls/toggle_layout_option.dart
Brazol: I don't like then also but they are used across whole SDK so it would require some other cleanup task

esarbanis: Agreed, I can do that, it would be a fast win for me 🙂

dogfooding/lib/screens/lobby_screen.dart
esarbanis: OK I got it now, thanks!

esarbanis (approved):
LGTM

PR #585, opened 2024-02-20
feat: Extended call stats event/stream added processed publisher/subscriber stats in CallState

Brazol:


esarbanis (approved):
A couple of things we noted:
- It may be better to expose 2 streams, one for raw data and one for processed structured data. Processing has a performance impact, and for something that it is called every 2 seconds this might add up.
- `SharedEmitter` is not really needed, as `Stream` or `StreamController` does exactly the same thing and it comes from the SDK.

Other than that LGTM

PR #586, opened 2024-02-21
feat(dogfooding): Settings menu button with reactions and call stats screen

Brazol:
Screenshot 2024-02-21 at 13 15 54
Screenshot 2024-02-21 at 13 11 03

### DIFF
--- a/dogfooding/lib/screens/call_stats_screen.dart
+++ b/dogfooding/lib/screens/call_stats_screen.dart
@@ -1,0 +1,291 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_dogfooding/app/user_auth_controller.dart';
+import 'package:flutter_dogfooding/di/injector.dart';
+import 'package:flutter_dogfooding/screens/stats_latency_chart.dart';
+import 'package:flutter_dogfooding/theme/app_palette.dart';
+import 'package:stream_video_flutter/stream_video_flutter.dart';
+
+class CallStatsScreen extends StatefulWidget {
+  const CallStatsScreen({
+    super.key,
+    required this.call,
+  });
+
+  final Call call;
+
+  @override
+  State<CallStatsScreen> createState() => _CallStatsScreenState();
+}
+
+class _CallStatsScreenState extends State<CallStatsScreen> {
+  final _userAuthController = locator.get<UserAuthController>();
+
+  @override
+  Widget build(BuildContext context) {
+    final streamVideoTheme = StreamVideoTheme.of(context);
+    final textTheme = streamVideoTheme.textTheme;
+    final currentUser = _userAuthController.currentUser;
+
+    return StreamBuilder<CallState>(
+        stream: widget.call.state.asStream(),
+        builder: (context, snapshot) {
+          final state = snapshot.data;
+          final subscriberBitrate = state?.subscriberStats?.bitrateKbps;
+          final publisherBitrate = state?.publisherStats?.bitrateKbps;
+
+          return Scaffold(
+            appBar: AppBar(
+              title: Text(
+                'Stats',
+                style: textTheme.title3.apply(color: Colors.white),
+              ),
+              centerTitle: true,
+              backgroundColor: Theme.of(context).scaffoldBackgroundColor,
+              actions: [
+                IconButton(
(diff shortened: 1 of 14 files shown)


packages/stream_video_flutter/lib/src/call_participants/call_participant.dart
```diff
@@ -202,7 +202,7 @@ class StreamCallParticipant extends StatelessWidget {
                   ),
                   if (participant.reaction != null)
                     Align(
-                      alignment: Alignment.topRight,
+                      alignment: Alignment.topCenter,
```
esarbanis, line 205: Since this is in our SDK shouldn't we have it configurable? 🤔

dogfooding/lib/screens/stats_latency_chart.dart
```diff
@@ -0,0 +1,121 @@
+import 'dart:math';
+
+import 'package:flutter/material.dart';
+import 'package:fl_chart/fl_chart.dart';
+import 'package:flutter_dogfooding/theme/app_palette.dart';
+
+class StatsLatencyChart extends StatefulWidget {
```
esarbanis, line 7: Why is this `Stateful`?

dogfooding/lib/screens/call_stats_screen.dart
```diff
@@ -0,0 +1,291 @@
...
+                    StatsItem(
+                      title: 'Receive bitrate',
+                      value: subscriberBitrate == null
+                          ? '--'
+                          : '${state.subscriberStats?.bitrateKbps} Kbps',
+                    ),
+                    // LatencyOrJitterItem(
+                    //   title: 'Publish resolution',
+                    //   value: state.publisherStats?.resolution,
+                    // ),
```
esarbanis, line 164: Maybe remove? 🙂

dogfooding/lib/screens/call_stats_screen.dart
```diff
@@ -0,0 +1,291 @@
...
+                    color: Colors.white,
+                  ),
+                  onPressed: () {
+                    Navigator.of(context).pop();
+                  },
+                ),
+              ],
+            ),
+            // StatItem(title = "Region", value = statsReport?.local?.sfu)
+            // StatItem(title = "Receive resolution", value = subscriberResolution)
```
esarbanis, line 57: Maybe remove? 🙂

dogfooding/lib/screens/call_stats_screen.dart
```diff
@@ -0,0 +1,291 @@
...
+  });
+
+  final Call call;
+
+  @override
+  State<CallStatsScreen> createState() => _CallStatsScreenState();
+}
+
+class _CallStatsScreenState extends State<CallStatsScreen> {
+  final _userAuthController = locator.get<UserAuthController>();
```
esarbanis, line 21: Shouldn't we `dispose` this?

esarbanis (changes requested):

packages/stream_video_flutter/lib/src/call_participants/call_participant.dart
Brazol: Yes true, I would leave it for now and addressed it in separate task because I think we need to redo the reaction displaying component. There should be a builder callback for it so users can customize it and we should make it disapear after some time.

dogfooding/lib/screens/stats_latency_chart.dart
Brazol: ah sorry, forgot to switch back to Stateless after changing implementation of it

dogfooding/lib/screens/call_stats_screen.dart
Brazol: it's a singleton so not here, it will be disposed when locator is cleaned up

esarbanis: But if it is a singleton, we do not really have any state to track here, shouldn't we consider switching to `StatelessWidget`?

Brazol: you are right, fixed

esarbanis (approved):
LGTM

PR #587, opened 2024-02-21
feat(llc): add static typing to call types

esarbanis:


Brazol:
Woudn't it be easier to handle it like this, or is there something I don't see?

```
class CallType {
  final String value;

  CallType._(this.value);

  // ignore: non_constant_identifier_names
  factory CallType.Default() => CallType._('default');
  factory CallType.audio() => CallType._('audio');
  factory CallType.custom(String value) => CallType._(value);
}
```

esarbanis:
I see what you are suggesting, how would we convert a `String` to `CallType` in case of `cid` splitting?

Brazol:
Wouldn't it be easier to handle it with a single CallType class instead of having to register a factory. Something like this:

```
class CallType {
  final String value;

  CallType._(this.value);

  // ignore: non_constant_identifier_names
  factory CallType.defaultCall() => CallType._('default');
  factory CallType.audioRoom() => CallType._('audioRoom');
  factory CallType.liveStream() => CallType._('liveStream');
  factory CallType.development() => CallType._('development');

  factory CallType.custom(String value) => CallType._(value);

  factory CallType.fromAlias(String alias) {
    switch (alias) {
      case 'default':
        return CallType.defaultCall();
      case 'audioRoom':
        return CallType.audioRoom();
      case 'liveStream':
        return CallType.liveStream();
      case 'development':
        return CallType.development();
      default:
        return CallType.custom(alias);
    }
  }
}
```

Or am I missing something? :)

esarbanis:
The point was to protect users from passing unknown types say they have setup `custom` but they send `custom1:1` . `CallType.fromAlias('custom1')` will parse it as is.

docusaurus/docs/Flutter/03-core-concepts/05-call-types.mdx
```diff
@@ -6,6 +6,27 @@ description: How Call Types control features and permissions
...
+There are 4 built-in call types, `default`, `audio_room`, `livestream` and `development`.
+
+You can also create your own types by calling `StreamCallType.custom`.
+  
+```dart
+final call = client.makeCall(callType: StreamCallType.custom("custom"), "123")
+```
+
+:::note
+  When using a custom call type, you need to make sure that the server is configured to handle the custom call type.
```
Brazol, line 26: maybe we should add a link or some kind of pointer to the Stream dashboard where user can do this?
so something like: https://dashboard.getstream.io/app/{you_app_id}/video/call-types/new

dogfooding/ios/Runner.xcodeproj/project.pbxproj
Brazol, line 1: should this be part of the PR?

packages/stream_video/lib/src/call/call_type.dart
```diff
@@ -0,0 +1,34 @@
...
+/// - [StreamCallType.Default()] (default)
+/// - [StreamCallType.liveStream] (livestream)
+/// - [StreamCallType.development] (development)
+/// - [StreamCallType.audioRoom] (audio_room)
+class StreamCallType {
+  const StreamCallType._(this.value);
+
+  final String value;
+
+  factory StreamCallType.Default() => const StreamCallType._('default');
```
Brazol, line 13: I know that we can't user 'default' starting with the lowercase letter but maybe we could use `defaultCall`? It breaks the pattern where factory constructor corresponds 1-1 to the string name of the type but Uppercase letter breaks the dart convention and looks bad.

docusaurus/docs/Flutter/03-core-concepts/05-call-types.mdx
esarbanis: I want to find a link to documentation for this ... will ask around.

dogfooding/ios/Runner/Info.plist
```diff
@@ -44,7 +44,7 @@
 		</dict>
 	</array>
 	<key>CFBundleVersion</key>
-	<string>$(FLUTTER_BUILD_NUMBER)</string>
+	<string>64</string>
```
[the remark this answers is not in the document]
esarbanis: Oops 🙂

packages/stream_video/lib/src/call/call_type.dart
esarbanis: It is too verbose ... how about `StreamCallType.default_()`?

esarbanis: Or just:
```dart
  factory StreamCallType() => const StreamCallType._('default');
 ```

File: packages/stream_video/CHANGELOG.md [hunk 4]
```diff
@@ -14,6 +14,7 @@
 ✅ Added
 
 * Added `custom` field to `CallParticipantState` with custom user data.
+* Added `CallType` to statically track the type of call (audio, video, screen share).
```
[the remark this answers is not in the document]
esarbanis: It is not breaking, I deprecated the `String` version, you have to pass either one or both.

docusaurus/docs/Flutter/03-core-concepts/05-call-types.mdx
esarbanis: I embedded [this page](https://github.com/GetStream/stream-chat-docusaurus-cli/blob/production/shared/video/_call-types.m) according to the other SDKs

Brazol (approved):
LGTM

PR #589, opened 2024-02-23
feat(dogfooding): env switcher + dynamic link support

Brazol:
Environment switcher on login screen + dynamic link support for all environments. 

https://github.com/GetStream/stream-video-flutter/assets/5622717/6f16d23e-44b3-4f74-bad5-91a53be8f8d7

### DIFF
--- a/dogfooding/lib/core/repos/token_service.dart
+++ b/dogfooding/lib/core/repos/token_service.dart
@@ -4,6 +4,38 @@ import 'dart:convert';
 // 📦 Package imports:
 import 'package:http/http.dart' as http;
 
+enum Environment {
+  pronto(
+    'Pronto',
+    aliases: ['stream-calls-dogfood'],
+  ),
+  demo(
+    'Demo',
+    aliases: [''],
+  ),
+  staging(
+    'Staging',
+  );
+
+  final String displayName;
+  final List<String> aliases;
+
+  const Environment(
+    this.displayName, {
+    this.aliases = const [],
+  });
+
+  factory Environment.fromHost(String host) {
+    final hostParts = host.split('.');
+    final String envAlias = hostParts.length < 2 ? '' : hostParts[0];
+
+    return Environment.values.firstWhere(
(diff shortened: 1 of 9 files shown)


dogfooding/lib/core/repos/token_service.dart
```diff
@@ -4,6 +4,38 @@ import 'dart:convert';
...
 import 'package:http/http.dart' as http;
 
+enum Environment {
+  pronto(
+    'Pronto',
+    aliases: ['stream-calls-dogfood'],
+  ),
+  demo(
+    'Demo',
+    aliases: [''],
```
esarbanis, line 14: Is this needed if so don't we have to put it in `staging` as well?

dogfooding/lib/screens/login_screen.dart
```diff
@@ -237,3 +246,104 @@ String randomId({int size = 21}) {
...
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Padding(
+      padding: const EdgeInsets.all(8.0),
+      child: Row(
+        children: [
+          Text(
+            'Current environment: ${selectedEnvironment.displayName}',
```
esarbanis, line 278: I think this is too verbose and feels out of place with the rest of the polished UI. How about making it a banner or a ribbon (like the debug one) with just the name of the environment?

dogfooding/lib/core/repos/token_service.dart
Brazol: Maybe it should be called subdomains instead of aliases. For demo env it just says that getstream.io without subdomain will map to it

esarbanis: Then why don't we put `aliases: ['staging']` is the `Environment.staging` constructor?

Brazol: oh yes thats why I named it alias :D because I'm matching the env when subdomain equals to the name of the env (demo, pronto, staging) or matches one of the aliases. So `statging` is not needed because its the name

PR #592, opened 2024-02-29
fix: dogfooding env handling

Brazol:
- Fixes environment handling in Dogfooding
- Disables back gestures on the call screen
- Fixes the latency chart axis labels


esarbanis (approved):
LGTM

PR #595, opened 2024-03-03
fix call identifier format on handleVoipPushNotification function

iJack93:
This pull request addresses a critical issue where the call identifier, formatted as "vjWxobiv0xhkKQ6ze6MfSSwY6BL2", was inadvertently used in place of the new UUID (callId) for the 'showIncomingCall' and 'showMissedCall' methods. This substitution resulted in an invalid UUID format that is incompatible with platform channel methods such as UUID(uuidString: data.uuid) in Swift: this led to the creation of a nil object and triggered a crash due to the forced unwrapping of a non-nullable value.

### DIFF
--- a/packages/stream_video/lib/src/stream_video.dart
+++ b/packages/stream_video/lib/src/stream_video.dart
@@ -584,13 +584,14 @@ class StreamVideo {
     final callCid = payload['call_cid'] as String?;
     if (callCid == null) return false;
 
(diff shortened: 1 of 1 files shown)


esarbanis:
Thank you for your effort you put into this PR,  unfortunately we were already working on a fix for this ... I am closing this PR in favour of #599

PR #596, opened 2024-03-04
chore(repo): distribute dogfooding only when there are changes to it

esarbanis:
As described [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore)


Brazol (approved):
LGTM

PR #601, opened 2024-03-05
feat(stats): local call stats added + stats documentation

Brazol:


docusaurus/docs/Flutter/03-core-concepts/08-call-statistics.mdx
```diff
@@ -0,0 +1,41 @@
+---
+title: Call Statistics
+slug: /call-statistics
+description: How to listen and leverage call statistics
+---
+
+There are two sources of statistics you can use to monitor the performance of your calls:
+- `call.stats` stream that provides real-time webRTC statistics
```
esarbanis, line 8: Maybe capitalize `call` to be clear we are referring to the class

docusaurus/docs/Flutter/03-core-concepts/08-call-statistics.mdx
```diff
@@ -0,0 +1,41 @@
...
+- `CallState` properties that contain some of the webRTC statistics processed and more useful for display purposes
```
esarbanis, line 9: It would be useful to add a link to the code so that they see the structure

dogfooding/lib/screens/call_screen.dart
```diff
@@ -124,6 +124,7 @@ class _CallScreenState extends State<CallScreen> {
 
   @override
   Widget build(BuildContext context) {
+    // ignore: deprecated_member_use
     return WillPopScope(
```
esarbanis, line 128: Maybe we should get rid of this deprecation otherwise we will not be able to support new Flutter version (this comes from experience 🙂 )

Brazol: It's tricky because the suggested replacement PopScope doesn't offer the same functionality. Lets leave it for now :)

PR #607, opened 2024-03-13
feat(llc): add stats sending

esarbanis:


packages/stream_video/lib/src/call/call.dart
```diff
@@ -48,6 +49,7 @@ const _idSessionEvents = 4;
 const _idSessionStats = 5;
 const _idConnect = 6;
 const _idAwait = 7;
+const _idSessionStatsSent = 8;
```
Brazol, line 52: I think it's not used anywhere?

packages/stream_video/lib/src/call/session/call_session.dart
```diff
@@ -602,6 +604,23 @@ class CallSession extends Disposable {
         raw: rtcRawStats,
       ),
     );
+
+    sfuClient.sendStats(
```
Brazol, line 608: Is is ok that we will send publisherStats and subsciberStats separately? Shouldn't we wait to have both and send it together? It also means that we won't respect `reportingIntervalMs` fully since we will send two reports every interval.

Brazol (approved):
LGTM

PR #613, opened 2024-03-22
fix(llc): fixing call participants creation when receving call details

Brazol:
Previously wherever we get a call details we updated the CallState with participants created from member list. First problem was that it was losing the state of CallParticipant which contains publishedTracks. Second is that it should iterate session participats and not members collection.


deven98 (approved):
LGTM

PR #614, opened 2024-03-22
feat: Picture in picture support for Android

Brazol:


packages/stream_video_flutter/lib/stream_video_flutter_method_channel.dart
```diff
@@ -103,4 +103,17 @@ class MethodChannelStreamVideoFlutter extends StreamVideoFlutterPlatform {
       },
     );
   }
+
+  @override
+  Future<void> togglePictureInPicture({required bool enable}) async {
```
esarbanis, line 108: Toggling implies that it is going to switch based on the current state, but since we are passing the state, it is no longer a toggle. Maybe setPictureInPicture?

esarbanis (approved):
LGTM

PR #641, opened 2024-04-09
feat(llc): added mirror parameter to CameraConstraints

Brazol:


packages/stream_video/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Unreleased
+✅ Added
+
+* `mirror` parameter to `CameraConstraints`, that is used to determine if the camera for the given track should be mirrored or not. When set as null the default is true when f`acingMode` is set to `user` and false when facingMode is set to `environment`.
```
deven98, line 4:
```suggestion
* Added `mirror` parameter to `CameraConstraints`, which determines whether the camera for the given track should be mirrored or not. When set as null the default is true when `facingMode` is set to `user` and false when facingMode is set to `environment`.
```

packages/stream_video/lib/src/webrtc/media/camera_constraints.dart
```diff
@@ -4,29 +4,34 @@ import 'constraints/facing_mode.dart';
 import 'video_constraints.dart';
 
 /// Options used when creating a video track that captures the camera.
+/// [mirror]: Whether the video should be mirrored. If set to null, the default is true when facingMode is user and false when facingMode is environment.
```
deven98, line 7: Should we add docs to the declaration of the mirror variable?

packages/stream_video/lib/src/webrtc/media/camera_constraints.dart
```diff
@@ -52,11 +57,13 @@ class CameraConstraints extends VideoConstraints {
     FacingMode? facingMode,
     String? deviceId,
     double? maxFrameRate,
+    bool? Function()? mirror,
```
deven98, line 60: Why a function here?

Brazol: It's a nullable field so you can't set it to null using copyWith otherwise. How do we handle it in other places?

deven98: Why not make it non-nullable then?

deven98: Set it to false by default?

Brazol: Because by default it should be true for front camera and false for back camera. So it has 3 states. True, false, default (null). We could make it an enum

deven98 (approved):
LGTM

PR #643, opened 2024-04-09
feat(push_notification): add capacity to set has_video flag for Callkit notification

TinhHuynh:
This PR allows the client application to determine which Callkit notification UI is shown (audio or video) using callerCustomizationCallback.

### DIFF
--- a/dogfooding/pubspec.yaml
+++ b/dogfooding/pubspec.yaml
@@ -29,8 +29,8 @@ dependencies:
   share_plus: ^7.1.0
   shared_preferences: ^2.2.0
   stream_chat_flutter: ^7.1.0
-  stream_video_flutter: ^0.3.5
-  stream_video_push_notification: ^0.3.5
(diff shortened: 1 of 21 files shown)


Brazol:
Hello @TinhHuynh, 

I wanted to thank you for your contribution. We are aware of the issue and plan to address it. However, the solution you suggested, which we also considered before, may not work as expected. Specifically, `callerCustomizationCallback` is not called for iOS in the background or terminated app state. This means that while it may work fine on Android and for foregrounded apps on iOS, it may not be a complete solution. 

Thank you for your understanding.

TinhHuynh:
Hi @Brazol
Thanks for the feedback. Yes, I know the solution does not work in iOS when the app is in the background or terminated state. Given the state of the Flutter SDK, that's the best I can contribute :), the complete solution may need Stream backend which I cannot access.
That's glad you have a plan to address the issue. Again, thank you for your time 🙏

PR #649, opened 2024-04-11
fix: fixing screen sharing and switching screen sharing type on iOS

Brazol:


packages/stream_video_flutter/lib/src/call_screen/call_content/call_content.dart
```diff
@@ -161,7 +161,7 @@ class _StreamCallContentState extends State<StreamCallContent>
   Widget build(BuildContext context) {
     final theme = StreamVideoTheme.of(context);
 
-    if (_isPictureInPictureModeOn) {
+    if (_isPictureInPictureModeOn && CurrentPlatform.isAndroid) {
```
deven98, line 164: Doesn't this restrict screensharing to iOS?

Brazol: It's for Picture in Picture :) It's not connected to the whole PR it's just something I noticed and fixed in the meantime.

deven98 (approved):
LGTM

PR #652, opened 2024-04-12
fix: set call notifications to low priority with no sound

Brazol:


deven98 (approved):
LGTM - pending testing from customer

PR #664, opened 2024-05-07
fix: fixing iOS speakerphone initialization

Brazol:


packages/stream_video/lib/src/webrtc/rtc_manager.dart
```diff
@@ -900,3 +917,7 @@ extension on RtcLocalTrack<VideoConstraints> {
     return dimension;
   }
 }
+
+extension on String {
```
deven98, line 921: Can we just create a String extensions file?

deven98 (approved):
LGTM

PR #666, opened 2024-05-07
fix: delayed subscriber offer handling

Brazol:


packages/stream_video/lib/src/call/session/call_session.dart
```diff
@@ -106,8 +109,25 @@ class CallSession extends Disposable {
...
+      await _eventsSubscription?.cancel();
+      await _rtcManagerSubject?.close();
+
+      _rtcManagerSubject = BehaviorSubject();
+
+      // Buffer sfu events until rtc manager is set
+      final bufferedStream =
+          sfuWS.events.asStream().buffer(_rtcManagerSubject!);
+
+      // Handle buffored events and then listen to sfu events as normal
```
deven98, line 122:
```suggestion
      // Handle buffered events and then listen to sfu events as normal
```

deven98 (approved):
LGTM

PR #681, opened 2024-05-17
fix: fixed screensharing persistence and track issue

deven98:
Fixes screen sharing view persisting beyond stopping sharing as well as the track not showing in the view.


deven98:
IMP: Before merging - the key solutions for the screen sharing issue are the keys for the video renderer and the logic for when to display the screen share screen.

The PR also changes keys for stream call participant. Check with backend - if session IDs are unique to every user, maybe it is better to keep it as it is. If a session ID is shared with multiple users, then either use user ID or combine user ID + session ID.

Brazol (approved):
LGTM

PR #685, opened 2024-05-22
fix: full reconnect flow

deven98:
Fixes full reconnect issue when multiple users showed up and tracks disappeared.


Brazol (approved):
LGTM

PR #699, opened 2024-06-14
fix: fixed accepting ringing during call

Brazol:
Fixed accepting incoming call during another call (or outgoing call)


packages/stream_video/lib/src/core/client_state.dart
```diff
@@ -21,14 +20,18 @@ abstract class ClientState {
   StateEmitter<Call?> get activeCall;
 
   /// Emits when a call was created by another user with ringing set as True.
-  StateEmitter<Call> get incomingCall;
+  StateEmitter<Call?> get incomingCall;
+
+  /// Emits when a call was created by another user with ringing set as True.
```
kanat, line 25:
```suggestion
  /// Emits when a call was created by current user with ringing set as True.
```

kanat (approved):
LGTM

Issue #711, opened 2024-07-18
Export widgets


joaopinacio:
I need to use the FloatingViewAlignment class to change the initial alignment of the floating Video on the screen, but I noticed that you didn't create a file to export this specific Enum.

Could you help me find it or if it doesn't exist, add it in the next version?

```
...
Theme(
  data: ThemeData(
    extensions: [
      darkAppTheme.copyWith(
        localVideoTheme: darkAppTheme.localVideoTheme.copyWith(
          initialAlignment: FloatingViewAlignment.topLeft,
        ),
      ),
    ],
  ),
  child: StreamCallContainer(
    call: widget.call,
    enablePictureInPicture: true,
...
```

joaopinacio:
any updates?

Brazol:
Hi @joaopinacio,

Apologies for the wait, we already prepared the [PR](https://github.com/GetStream/stream-video-flutter/pull/723) with the change. It will be part of the next release.

PR #723, opened 2024-08-01
fix: export floating view alignment enum

Brazol:
Solves https://github.com/GetStream/stream-video-flutter/issues/711


deven98 (approved):
LGTM

Issue #706, opened 2024-06-25
Retrieve records from an ended call


qu0cquyen:
Hi GetStream team, first of all, thank you for your awesome library. 

I'm wondering if there is any ways that can query recorded calls without having an ongoing active call?

Like React library, where `queryCalls` will return an object `Call`. and this object `Call` can access `Recorded calls`  without having an ongoing connected active call.

Brazol:
Hi @qu0cquyen we have added `listRecordings()` method to the StreamVideo class so that you can now get list of recordings by call ID. It's now part of the latest release. I hope this resolves your issue.

PR #724, opened 2024-08-01
chore(llc): added listRecordings method to StreamVideo class

Brazol:
Solves https://github.com/GetStream/stream-video-flutter/issues/706


deven98 (approved):
LGTM

PR #747, opened 2024-09-04
fix: setting lockscreen visibility to public

Brazol:
Setting a lockscreen visibility for a call ringing notrification to public.


deven98 (approved):
LGTM

PR #748, opened 2024-09-04
fix: fixing on leave call callback

Brazol:
Fixes the onLeaveCallTap callback in `StreamCallContent` that was assigned to wrong widget


deven98 (approved):
LGTM

PR #768, opened 2024-09-23
feature(llc): Reconnects v2

Brazol:
**Overview**

Implements the Reconnects v2 flow. The implementation doesn't change the public API (not breaking); internally, it does change many things. The Reconnects v2 consists of three reconnect strategies:

- **FAST**: typically used when switching networks (wifi->5g or similar). This strategy takes the SFU-provided threshold, which normally is set to 25 seconds. The client attempts to connect to the same SFU and performs an ICE restart on the existing peer connections.
- **REJOIN**: used when FAST isn't applicable, failed or the SFU told us to do it. Drops all existing peer connections, closes the web socket to the SFU, fetches new credentials from the Coordinator, and finally establishes a connection to the SFU returned by the coordinator. The important detail here is, that we rejoin with a new session_id parameter. This strategy is analogous to a browser tab refresh.
- **MIGRATE**: when the SFU reaches its operational capacity or when it gets shut down due to maintenance or upgrade, it will ask connected clients to migrate away to another, healthy SFU. The migration is marked as complete, once the `participantMigrationComplete` event is emitted. If the migration fails, the client automatically attempts REJOIN.
- **DISCONNECT**: The SFU politely asks you to leave the call


deven98 (approved):
LGTM

PR #771, opened 2024-09-24
fix: migrating to internet connection checker plus

Brazol:
Due to [license issues](https://github.com/github/dmca/blob/master/2024/09/2024-09-04-internet-connection-checker-plus.md) `internet_connection_checker` version was reverted. This PR migrates to `internet_connection_checker_plus` instead.


sachaarbonel (approved):
LGTM

PR #777, opened 2024-10-15
fix(push): send createDevice request only once per token change

Brazol:


packages/stream_video_push_notification/lib/src/stream_video_push_notification.dart
```diff
@@ -174,13 +178,24 @@ class StreamVideoPushNotificationManager implements PushNotificationManager {
       return;
     }
 
-    void registerDevice(String token, bool isVoIP) {
-      _client.createDevice(
+    void registerDevice(String token, bool isVoIP) async {
+      final tokenKey = isVoIP ? 'userDeviceTokenVoIP' : 'userDeviceToken';
```
deven98, line 182: Can we change this key to something unique and identifies this as a stream key? I feel like these are names that someone could use by accident in their own app.

packages/stream_video_push_notification/lib/src/stream_video_push_notification.dart
```diff
@@ -197,6 +212,11 @@ class StreamVideoPushNotificationManager implements PushNotificationManager {
             .listen((token) => registerDevice(token, true)));
   }
 
+  Future<void> removedStoredTokens() async {
+    await _sharedPreferences.remove('userDeviceToken');
```
deven98, line 216: Additionally can we store these keys as compile time constants somewhere else?

packages/stream_video_push_notification/pubspec.yaml
```diff
@@ -24,7 +24,8 @@ dependencies:
   rxdart: ^0.28.0
   stream_video: ^0.5.5
   uuid: ^4.2.1
-
+  shared_preferences: ^2.3.2
+
```
deven98, line 28: Can you please add a changelog entry for everything?

deven98 (changes requested):

packages/stream_video_push_notification/lib/src/stream_video_push_notification.dart
Brazol: good point

PR #789, opened 2024-10-30
feat(llc): support target resolution setting

Brazol:
### 🎯 Goal

Support the target resolution setting configured on the dashboard for both camera and screen sharing.


deven98 (approved):
LGTM

PR #790, opened 2024-10-31
feat(llc): Added participants count in CallState

Brazol:
**Overview**

Handlers for the `call.session_participant_count_updated` event.
The Coordinator will occasionally deliver this event to all clients, and it will carry the most recent state of:

`anonymous_participant_count` and `participant_count_by_role`

**Implementation notes**

`watch` parameter is now added to `Call.get()` and `Call.getOrCreate()` methods. When set to `true` the coordinator events will be listened to even before joining the call.

The `CallState` is extended with `participantCount` and `anonymousParticipantCount` fields.
This values from the `call.session_participant_count_updated` event are used to update those fields but only when the call status isn't JOINED. Once the call is JOINED, these values will be populated from the SFU heart checks.

### DIFF
--- a/packages/stream_video/lib/src/call/call.dart
+++ b/packages/stream_video/lib/src/call/call.dart
@@ -255,6 +255,9 @@ class Call {
     return _connectOptionsOverride ?? _connectOptions;
   }
 
+  /// It is better to pass the [connectOptions] to [join] method,
+  /// setting it directly has to be done carefully. Depending on the moment in the call lifecycle,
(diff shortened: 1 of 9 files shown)


packages/stream_video/lib/src/call/call.dart
```diff
@@ -451,6 +466,7 @@ class Call {
     return result;
   }
 
+  /// Rejects the incomming call.
```
deven98, line 469:
```suggestion
  /// Rejects the incoming call.
```

PR #791, opened 2024-11-05
feat(llc): manual quality selection

Brazol:
Adds APIs to manually override incoming video quality:

`await call.setPreferredIncomingVideoResolution(VideoResolution(width: 640, height: 480), sessionIds: [ids]);`

Specify preference for incoming video resolution. The preference will be matched as close as possible, but the actual resolution will depend on the video source quality and client network conditions. It will enable incoming video if it is previously disabled.

Passing `null` resolution clears previously set preferences.

Optionally specify session ids of the participants this preference has effect on. Affects all participants by default.

`call.setIncomingVideoEnabled(false)`

Enables or disables incoming video from all remote call participants, and clears any preference for preferred resolution.

https://github.com/user-attachments/assets/cacf4bea-03b4-43c0-9815-5533b71a05e6


deven98 (approved):
LGTM

PR #793, opened 2024-11-07
feat: call quality rating

Brazol:
Implemented a call quality feedback feature in the dogfooding app and updated the necessary API in the Call class.

https://github.com/user-attachments/assets/4522c607-b990-4d44-afcd-656df1cc7d54


deven98 (approved):
LGTM

PR #801, opened 2024-11-28
chore: removed docusaurus

Brazol:
Removed Docusaurus since docs are now moved to separate repository.


xsahil03x (approved):
Nice, we can also remove the `.styles` folder.

xsahil03x (approved):

PR #815, opened 2024-12-30
chore: Updated minimum Flutter version to 3.24.5

Brazol:


deven98 (approved):
LGTM

dogfooding/pubspec.yaml
```diff
@@ -5,7 +5,7 @@ description: Flutter Dogfooding App to showcase Video SDK.
 
 environment:
   sdk: ">=3.4.0 <4.0.0"
-  flutter: ">=3.22.0"
+  flutter: ">=3.24.5"
```
xsahil03x, line 8: the new `color.withValues` is not available in the latest `3.27` so maybe set it as the min version for dogfooding.

dogfooding/lib/screens/stats_latency_chart.dart
```diff
@@ -96,14 +96,16 @@ class StatsLatencyChart extends StatelessWidget {
                   colors: [
                     ColorTween(
                             begin: AppColorPalette.appGreen,
-                            end: AppColorPalette.appGreen.withOpacity(0.5))
+                            end:
+                                AppColorPalette.appGreen.withValues(alpha: 0.5))
                         .lerp(0.2)!
-                        .withOpacity(0.1),
+                        .withValues(alpha: 0.1),
```
xsahil03x, line 102: Also, the alpha now ranges from (0 - 255), so we also have to multiply it by 255
```suggestion
			.withValues(alpha: 255.0 * 0.1)
```

xsahil03x: I guess, it's better to ignore the deprecation at this stage and we can replace it in the next major upgrade.

Brazol: I also just now figured out that it will fail for 3.24, so I reverted back to `withOpacity()` 😅 
But you don't have to multiply by 255, `withValues(alpha: )` works with doubles just like withOpacity() and not like withAlpha() - tested this.

dogfooding/pubspec.yaml
Brazol: You are right, reverted to `withOpacity()`

PR #824, opened 2025-01-13
fix(llc): Fix for ringing call rejection

Brazol:
In some cases `CallEnded` event is triggered instead of `CallDecline` when declining the ringing call by CallKit. In those cases call wasn't rejected properly.


deven98 (approved):
LGTM

PR #834, opened 2025-01-21
fix: fix for accepting second call when already in a call

Brazol:
Fixes the issue with accepting incoming call when another call is already active. Also fixes a dogfooding issue event subscriptions not being cleared properly when logging out.

### DIFF
--- a/packages/stream_video_push_notification/lib/src/stream_video_push_notification.dart
+++ b/packages/stream_video_push_notification/lib/src/stream_video_push_notification.dart
@@ -168,6 +168,7 @@ class StreamVideoPushNotificationManager implements PushNotificationManager {
             subscribeToEvents();
           } else if (event is ActionCallAccept ||
               event is ActionCallDecline ||
+              event is ActionCallTimeout ||
               event is ActionCallEnded) {
             _subscriptions.cancel(_idCallAccepted);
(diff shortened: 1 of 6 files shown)


packages/stream_video_push_notification/lib/src/stream_video_push_notification.dart
```diff
@@ -158,29 +157,23 @@ class StreamVideoPushNotificationManager implements PushNotificationManager {
...
-          _subscriptions.cancel(_idCallParticipantCount);
+          if (event is ActionCallIncoming) {
+            if (!client.isConnected) {
+              client.openConnection();
+            }
+
+            subscribeToEvents();
+          } else if (event is ActionCallAccept ||
+              event is ActionCallDecline ||
+              event is ActionCallEnded) {
```
xsahil03x, line 171: Is replacing `ActionCallTimeout` with `ActionCallEnded` intentional?

xsahil03x (approved):
LGTM, added a comment just to verify if that was intentional.

packages/stream_video_push_notification/lib/src/stream_video_push_notification.dart
Brazol: nice catch!! Both should be there - fixed

PR #840, opened 2025-01-27
fix(llc): Fixes for various small ringing issues

Brazol:
- Fixes for various ringing issues
- Added geofencing and broadcasting settings to call getOrCreate/update
- Fix for ending PiP on iOS when call disconnects


deven98 (approved):
LGTM

PR #841, opened 2025-01-28
fix(CallKit): fix for missing app icon in CallKit screen when app is terminated

Brazol:
The app icon asset name was not passed correctly to the CallKit configuration when a VoIP push was received while the app was terminated.


deven98 (approved):
LGTM

PR #862, opened 2025-02-20
chore: setup build flavors for dogfooding app

renefloor:
### 🎯 Goal

On Beta TestFlight builds we do want to have the option to switch environments, while on App Store builds we don't. 
Currently we have it based on kDebugMode, but this means that you can only switch environments if you build locally.

### 🛠 Implementation details

This adds 3 flavors, dev, beta and prod with their own app icons on android, iOS and macOS. Other platforms don't support flavors, even though it's still available in the dart code.
We can now base some features on the app flavor, instead of the debug mode. In this case I've only used it for the env switcher, but we could also use it for example for the [diagnostic stats](https://github.com/GetStream/stream-video-flutter/blob/d416079fcb028627637145a18269b62d671e027a/packages/stream_video_flutter/lib/src/call_screen/call_content/call_content.dart#L269-L275).

The Distribution workflow automatically builds using beta, but can be set to prod in a manual run.

### 🎨 UI Changes

No relevant UI changes

### 🧪 Testing

I tested this locally by running macos, android, iOS and web.
I also checked if CI builds the right flavor: https://github.com/GetStream/stream-video-flutter/actions/runs/13416648005/job/37478987057

#### General

### ☑️Reviewer Checklist
- [ ] Sample runs & works


Brazol (approved):
LGTM

PR #867, opened 2025-02-24
chore(llc): update X-Stream-Client header

Brazol:
Update to the X-Stream-Client header according to the [new specification](https://www.notion.so/stream-wiki/X-Stream-Client-values-17b6a5d7f9f6804987bbdfec570698c6).

### DIFF
--- a/packages/stream_video/lib/composed_version.dart
+++ b/packages/stream_video/lib/composed_version.dart
@@ -1,12 +1,16 @@
+import 'package:meta/meta.dart';
+
 import 'src/platform_detector/platform_detector.dart';
 import 'version.g.dart';
 
-/// Default user agent for all requests
-final String defaultUserAgent =
-    'stream-video-dart-client-${CurrentPlatform.name}';
+@internal
+// Details regarding app name, version, os and device. Is set during initialization of StreamVideo instance.
+String? clientVersionDetails;
+
+const String defaultUserAgent = 'stream-video-dart-v$packageVersion';
 
 /// Current package version
 const packageVersion = streamVideoVersion;
 
 /// Composed client version
-final streamClientVersion = '$defaultUserAgent-$packageVersion';
+final streamClientVersion = '$defaultUserAgent|$clientVersionDetails';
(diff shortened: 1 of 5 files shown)


packages/stream_video/lib/composed_version.dart
```diff
@@ -1,12 +1,16 @@
...
 import 'version.g.dart';
 
-/// Default user agent for all requests
-final String defaultUserAgent =
-    'stream-video-dart-client-${CurrentPlatform.name}';
+@internal
+// Details regarding app name, version, os and device. Is set during initialization of StreamVideo instance.
+String? clientVersionDetails;
+
+const String defaultUserAgent = 'stream-video-dart-v$packageVersion';
```
Brazol, line 10: Should we continue using `dart` here, or should we switch to `flutter`? We currently use `dart` because `stream_video` is a pure Dart package that is independent of Flutter. However, when we send call stats requests, we are using `stream-flutter` as an SDK. This creates inconsistency, and I'm uncertain about the impact this has on backend statistics and the purpose of this header.

CC: @martinmitrevski

martinmitrevski: If a customer uses only the low-level client, it should be "dart". If they also use the UI SDK, it should be "flutter". Otherwise, we won't have a way to distinguish the ones that use only the LLC. 
Stats are reported on a different place, not in nessy, so it won't cause inconsistencies.

Brazol: There is no easy way to accomplish this. Without adding a wrapper around the `StreamVideo` instance in `stream_video_flutter` and requiring integrators to use it, we cannot modify a variable used in the `stream_video` package. There isn't a guaranteed point of entry that will always be invoked simply by including the `stream_video_flutter` package in the app, which complicates things

I tried a few hacky ways, but those didn't work. Can we just stay with `dart` for the time being? Maybe in the future we will add such wrapper but I don't think adding it just for the header is a good idea.

renefloor (approved):

packages/stream_video/lib/composed_version.dart
Brazol: Together with @renefloor we decided to go with `flutter` for now as a sdk identifier in the header.

PR #869, opened 2025-02-26
fix: Always build both platforms on push event

renefloor:
### 🎯 Goal

Run workflow on push

### 🛠 Implementation details

I assumed `inputs.build_x` would be null or undefined, so that's why I checked for `!= false`, but apparently default is false for on push because here nothing runs: https://github.com/GetStream/stream-video-flutter/actions/runs/13544785120

So now I check if the event_name is 'push'.

#### General

### ☑️Reviewer Checklist
- [ ] Sample runs & works


Brazol (approved):
LGTM

PR #873, opened 2025-03-03
feat: Noise cancellation feature for Android

Brazol:


packages/stream_video_noise_cancellation/lib/stream_video_noise_cancellation_method_channel.dart
```diff
@@ -12,21 +13,41 @@ class MethodChannelStreamVideoNoiseCancellation
...
+    }
+
     return methodChannel.invokeMethod<void>('setEnabled', {'enabled': enabled});
   }
 
   @override
   Future<bool?> deviceSupportsAdvancedAudioProcessing() {
+    if (!CurrentPlatform.isAndroid && !CurrentPlatform.isIos) {
+      throw UnimplementedError(
+          'The current platform (${CurrentPlatform.name}) does not support audio processing.');
```
renefloor, line 48: I think here I would just return `false`. That way the other methods should also not be called any time. Maybe you could also return `false` in `isEnabled`. Because it's not just unsupported by the method channel, it's known to be not supported by the platform. The other methods I can agree on throwing an exception, but I think for the `deviceSupportsAdvancedAudioProcessing` it makes sense to just return false. Otherwise you call `deviceSupportsAdvancedAudioProcessing` and get `The current platform does not support audio processing.`, which is basically part of the question you are asking.

renefloor: Maybe we could keep it here as `UnimplementedError`, because it's the base method channel, but just return false in the `NoiseCancellationAudioProcessor` which is the public api.

> deviceSupportsAdvancedAudioProcessing != deviceSupportsNoiseCancellation 

I understand that not all iOS and android devices support AdvancedAudioProcessing, but on any other device (windows, macos, web) for sure it would be supported right? So in the `NoiseCancellationAudioProcessor` I would also return false for this method.

packages/stream_video_noise_cancellation/lib/noise_cancellation_audio_processor.dart
```diff
@@ -0,0 +1,72 @@
...
+        () => 'Noise cancellation is not supported on this platform: $e',
+      );
+    }
+  }
+
+  @override
+  Future<Result<bool>> isEnabled() async {
+    try {
+      await StreamVideoNoiseCancellation().isEnabled() ?? false;
+      return Result.success(true);
```
renefloor, line 21: You're always returning `true` here if you don't get an error.
```suggestion
      final result = await StreamVideoNoiseCancellation().isEnabled() ?? false;
      return Result.success(result);
```

packages/stream_video_noise_cancellation/lib/noise_cancellation_audio_processor.dart
```diff
@@ -0,0 +1,72 @@
...
+
+  NoiseCancellationAudioProcessor() {
+    try {
+      StreamVideoNoiseCancellation().registerProcessor();
+    } on UnimplementedError catch (e) {
+      _logger.w(
+        () => 'Noise cancellation is not supported on this platform: $e',
+      );
+    }
+  }
```
renefloor, line 15: You're not catching async errors here, because you don't await. I think it's better to put it in a different async method:

```suggestion
  NoiseCancellationAudioProcessor() {
    _init();
  }
  
  Future<void> _init() async {
    try {
      await StreamVideoNoiseCancellation().registerProcessor();
    } on UnimplementedError catch (e) {
      _logger.w(
        () => 'Noise cancellation is not supported on this platform: $e',
      );
    } catch (err, stackTrace) {
      _logger.e(
        () => err.toString(),
      );
    }
  }
```

I also wonder if we shouldn't make an async factory so the caller knows if/when the noise cancelation is initialized.

packages/stream_video_noise_cancellation/android/src/main/kotlin/io/getstream/video/flutter/stream_video_noise_cancellation/StreamVideoNoiseCancellationPlugin.kt
```diff
@@ -0,0 +1,129 @@
...
+
+  override fun onMethodCall(call: MethodCall, result: Result) {
+    Log.d(TAG, "Method call: ${call.method}")
+    when (call.method) {
+      "registerProcessor" -> {
+        try {
+          controller = NoiseCancellationAudioProcessingController(applicationContext)
+          FlutterWebRTCPlugin.sharedSingleton.setAudioProcessingFactoryProvider(controller)
+          Log.d(TAG, "Noise cancellation processor registered successfully")
+          result.success(null)
```
renefloor, line 40: Do we need a check if the controller is already set? Don't know, maybe dispose an existing controller or throw an error that you registered twice?

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -944,6 +946,24 @@ class StreamVideo extends Disposable {
...
+  Future<Result<None>> setAudioProcessingEnabled(bool enabled) async {
+    return await _options.audioProcessor?.setEnabled(enabled) ??
+        const Result.failure(VideoError(message: 'No audio processor found.'));
+  }
+
+  Future<Result<bool>> deviceSupportsAdvancedAudioProcessing() async {
+    return await _options.audioProcessor
+            ?.deviceSupportsAdvancedAudioProcessing() ??
+        const Result.failure(VideoError(message: 'No audio processor found.'));
+  }
```
renefloor, line 966: related to the other comments, I would just return false here, instead of error.

Do we also need a method `deviceSupportsNoiseCanceling` or something similar?

Probably good to document here what exactly is `AdvancedAudioProcessing`.

renefloor (approved):

PR #882, opened 2025-03-06
feat: Add audioLevels and activeSpeakers

renefloor:
FLU-44

### 🎯 Goal

Add audioLevels and active speakers to match the swift sdk.

### 🛠 Implementation details

activeSpeakers is very similar to `otherParticipants` and `localParticipant`, so not fancy.

The list of audioLevels is maintained when doing `copyWith(audioLevel:...)`, so this level is added to the list every time the level changes.

I had to remove `const` to keep the API clean.
We can keep it const by adding a private field like this:
```dart
  final List<double>? _audioLevels;
  /// List of the last 10 audio levels.
   List<double> get audioLevels => _audioLevels ?? [audioLevel];
```
But I don't think it's worth it.

### 🎨 UI Changes

No UI changes

#### General

### ☑️Reviewer Checklist
- [ ] Sample runs & works

### DIFF
--- a/packages/stream_video/lib/src/models/call_participant_state.dart
+++ b/packages/stream_video/lib/src/models/call_participant_state.dart
@@ -14,7 +14,7 @@ import 'viewport_visibility.dart';
 class CallParticipantState
     with EquatableMixin
     implements Comparable<CallParticipantState> {
-  const CallParticipantState({
+  CallParticipantState({
     required this.userId,
     required this.roles,
     required this.name,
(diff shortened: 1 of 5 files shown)


packages/stream_video/lib/src/models/call_participant_state.dart
```diff
@@ -45,7 +46,13 @@ class CallParticipantState
   final bool isLocal;
   final SfuConnectionQuality connectionQuality;
   final bool isOnline;
+
+  /// The audio level for the user.
```
Brazol, line 50:
```suggestion
  /// The latest audio level for the user.
```

packages/stream_video/lib/src/models/call_participant_state.dart
```diff
@@ -73,6 +80,14 @@ class CallParticipantState
     CallReaction? reaction,
     ViewportVisibility? viewportVisibility,
   }) {
+    final levels = audioLevels;
```
Brazol, line 83: I'm not sure `copyWith` is the correct place for this logic. We can't assume that it will always be used to update the state with latest data. It might be used to override the whole state or any other reason. We should add `audioLevels` as a `copyWith` parameter and move this logic to the place where new/latest audioLevel is actually updated. Probably `sfuUpdateAudioLevelChanged`method.

renefloor: Changed it: https://github.com/GetStream/stream-video-flutter/pull/882/commits/f9847e147c07b5e9bcdc060c88dbc9b3d56e45b0

Brazol (approved):
LGTM

PR #892, opened 2025-03-12
chore: dogfooding custom credentials

Brazol:


dogfooding/lib/core/repos/custom_environment_loader.dart
```diff
@@ -0,0 +1,29 @@
...
+///
+/// Replace the global variable `customEnvironmentLoader` with your own implementation.
+///
+/// ```dart
+/// final customEnvironmentLoader = CustomEnvironmentLoader(
+///     tokenLoader: (userId) => '{USER_TOKEN}', apiKey
+///     apiKey: '{API_KEY}');
+/// ```
+class CustomEnvironmentLoader {
+  final String Function(String) tokenLoader;
```
renefloor, line 14: Probably better to make this async, so you can download a token from a backend?
```suggestion
  final Future<String> Function(String) tokenLoader;
```

dogfooding/lib/core/repos/token_service.dart
```diff
@@ -66,13 +74,27 @@ class TokenResponse {
...
+
+  const TokenService({this.customEnvironmentLoader});
 
   Future<TokenResponse> loadToken({
     required String userId,
     required Environment environment,
     Duration? expiresIn,
   }) async {
+    if (environment == Environment.custom) {
+      if (customEnvironmentLoader == null) {
```
renefloor, line 87: I think I would drop `Environment.custom`, because now people have to change 2 places to use a custom environment. I would just replace these 2 `if`s with
```suggestion
    if (customEnvironmentLoader != null) {
```
And remove the new custom env.

dogfooding/lib/core/repos/token_service.dart
```diff
@@ -66,13 +74,27 @@ class TokenResponse {
...
+        throw Exception(
+            'Custom environment is not configured. Provide customEnvironmentLoader to AppInjector.init method to use it.');
+      }
+
+      return TokenResponse(
+        customEnvironmentLoader!.tokenLoader(userId),
+        customEnvironmentLoader!.apiKey,
+      );
```
renefloor, line 95: If you make tokenLoader async you need
```suggestion
      final userToken = await customEnvironmentLoader!.tokenLoader(userId);

      return TokenResponse(
        userToken,
        customEnvironmentLoader!.apiKey,
      );
```

dogfooding/lib/core/repos/custom_environment_loader.dart
Brazol: replaced with FutureOr 👍

dogfooding/lib/core/repos/token_service.dart
Brazol: Users don't need to change anything here. The custom environemnt will appear in an env switcher on the login page. So they can choose `custom` there and then their custom loader will be used. Otherwise they can still use our ends if they want to

dogfooding/README.md
```diff
@@ -83,23 +91,54 @@ flutter doctor -v
...
+
+This is crucial for testing direct calls, as you'll need to reference the correct user_id.
+
+## Custom Environment Setup
+
+The **Dogfooding** app is primarily for Stream employees to test the Video SDK internally. You can use one of our environments (prefereably `demo` which is most stable one) for testing, however, it is recommended to configure a custom environment if you have a **Stream app** created in the **Stream Dashboard**.
+
+With **Dogfooding** connected yo your own **Stream app** you will have access to full configuration as well as **Call Stats**, **Events Logs** etc.
+
+Additionaly **APN Push Provider** used for ringing on **iOS devices** is tied to the iOS app **bundle identifier**. Since you will have to change the identifier to run the Dogfooding app using your own Apple Developer account, ringing won't work on iOS.
```
renefloor, line 123:
```suggestion
Additionaly **APN Push Provider** used for ringing on **iOS devices** is tied to the iOS app **bundle identifier**. Since you will have to change the identifier to run the Dogfooding app using your own Apple Developer account, ringing won't work on iOS with the default environments.
```

renefloor, line 123:
```suggestion
Additionaly **APN Push Provider** used for ringing on **iOS devices** is tied to the iOS app **bundle identifier**. Since you will have to change the identifier to run the Dogfooding app using your own Apple Developer account, ringing won't work on iOS with the default environments.
```

PR #897, opened 2025-03-13
feat: Added pin/unpin for everyone method to Call

Brazol:
FLU-66

Added `setParticipantPinnedForEveryone()` method to `Call` object that allows for pinning given session to the top of the list for everyone in the call. It requires user to have `pin-for-everyone` permission set in the Dashboard. 

The `setParticipantPinned()` method is still available and as before will only pin the session locally.

### DIFF
--- a/packages/stream_video/lib/src/call/call.dart
+++ b/packages/stream_video/lib/src/call/call.dart
@@ -2141,27 +2141,38 @@ class Call {
     return result;
   }
 
+  /// Pins/unpins the given session to the top of the participants list.
+  /// The change is done locally and won't affect other participants.
   Future<Result<None>> setParticipantPinned({
     required String sessionId,
     required String userId,
     required bool pinned,
   }) async {
-    final result = await _session?.setParticipantPinned(
-          sessionId: sessionId,
-          userId: userId,
-          pinned: pinned,
-        ) ??
-        Result.error('Session is null');
+    _stateManager.setParticipantPinned(
+      sessionId: sessionId,
+      userId: userId,
+      pinned: pinned,
+    );
 
-    if (result.isSuccess) {
-      _stateManager.setParticipantPinned(
-        sessionId: sessionId,
-        userId: userId,
-        pinned: pinned,
-      );
-    }
+    return const Result.success(none);
+  }
 
-    return result;
+  /// Pins/unpins the given session to the top of the participants list for everyone in the call.
+  /// You can execute this method only if you have the `pin-for-everyone` capability.
+  Future<Result<None>> setParticipantPinnedForEveryone({
+    required String sessionId,
+    required String userId,
+    required bool pinned,
+  }) async {
+    return pinned
+        ? _permissionsManager.pinForEveryone(
+            userId: userId,
+            sessionId: sessionId,
+          )
+        : _permissionsManager.unpinForEveryone(
+            userId: userId,
+            sessionId: sessionId,
+          );
   }
 
   /// Starts the livestreaming of the call.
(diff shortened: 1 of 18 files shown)


packages/stream_video/lib/src/call/call.dart
```diff
@@ -2141,27 +2141,38 @@ class Call {
...
 
-    if (result.isSuccess) {
-      _stateManager.setParticipantPinned(
-        sessionId: sessionId,
-        userId: userId,
-        pinned: pinned,
-      );
-    }
+    return const Result.success(none);
+  }
```
renefloor, line 2158: I wonder if we should also add a `setParticipantPinnedSync` and maybe deprecate this one. We can simplify it to just:
```dart
  void setParticipantPinnedSync({
    required String sessionId,
    required String userId,
    required bool pinned,
  }) {
    _stateManager.setParticipantPinned(
      sessionId: sessionId,
      userId: userId,
      pinned: pinned,
    );
  }
```

packages/stream_video/lib/src/call/call.dart
```diff
@@ -2141,27 +2141,38 @@ class Call {
...
 
-    return result;
+  /// Pins/unpins the given session to the top of the participants list for everyone in the call.
+  /// You can execute this method only if you have the `pin-for-everyone` capability.
```
renefloor, line 2161: I think this is a user specific capability right?
```suggestion
  /// Users can execute this method only if they have the `pin-for-everyone` capability.
```

packages/stream_video/lib/src/call/permissions/permissions_manager.dart
```diff
@@ -313,6 +313,42 @@ class PermissionsManager {
...
+      return Result.error('Cannot pin session for everyone (no permission)');
+    }
+
+    _logger.d(() => '[pinForEveryone] pinning $userId for everyone');
+
+    return coordinatorClient.videoPin(
+      callCid: callCid,
+      sessionId: sessionId,
+      userId: userId,
+    );
```
renefloor, line 331: Not for this PR, as it already happens in this file, but it feels weird to me that a permission manager actually executes the request. Either we could rename this permission manager, or change how this works. If I call `pinForEveryone` on a permission manager I would not expect that thing to happen directly.

packages/stream_video/lib/src/call/state/mixins/state_coordinator_mixin.dart
```diff
@@ -414,8 +414,8 @@ mixin StateCoordinatorMixin on StateNotifier<CallState> {
           audioLevel: e.audioLevel,
           isSpeaking: e.isSpeaking,
           isDominantSpeaker: e.isDominantSpeaker,
-          isPinned: e.isPinned,
           viewportVisibility: e.viewportVisibility,
+          pin: e.pin,
```
renefloor, line 418: I think this makes it a breaking change right? Or are these mixins not part of the public API or the core package?

packages/stream_video/lib/src/models/call_participant_state.dart
```diff
@@ -171,6 +199,7 @@ class CallParticipantState
         isSpeaking,
         isDominantSpeaker,
         isPinned,
+        pin,
```
renefloor, line 202: isPinned could be removed here.

packages/stream_video/lib/src/call/permissions/permissions_manager.dart
Brazol: Yep agreed. Tomasso also suggested that we shouldn't check permissions at all and leave it to the backend. Something to consider.

renefloor (approved):

Issue #887, opened 2025-03-08
onCallDisconnected function parameter


mlab817:
I think it would be very useful to have information on why the call disconnected here so the user can do something about it. We can provide options to user based on the call disconnection reason.

Brazol:
Hey @mlab817, thank you for your suggestion. It makes sense to include the reason for the disconnect in the callback. I added it in [this PR](https://github.com/GetStream/stream-video-flutter/pull/903), and it will be part of the next release.

PR #903, opened 2025-03-17
chore: added reason to onCallDisconnected callback

Brazol:
solves https://github.com/GetStream/stream-video-flutter/issues/887


packages/stream_video_flutter/lib/src/call_screen/call_container.dart
```diff
@@ -82,7 +82,7 @@ class StreamCallContainer extends StatefulWidget {
   final VoidCallback? onCancelCallTap;
 
   /// The action to perform when the call is disconnected. By default, it pops the current route.
-  final VoidCallback? onCallDisconnected;
+  final void Function(DisconnectReason)? onCallDisconnected;
```
renefloor, line 85: Instead of just returning a `DisconnectReason` I think it's good to return a new class, something like `DisconnectProperties` or something better. We can start with only adding the reason, but we would be able to add more properties without another breaking change.

Brazol: Done

renefloor (approved):

Issue #914, opened 2025-03-20
Inconsistent control behavior when user is joined using more than one device


mlab817:
version: 0.8.4

This is related to a support ticket I submitted.

---

How to Replicate

- User joins using 2 devices (both iOS devices in my case).
- Try mute/unmute audio
- Try mute/unmute video
- Try mute/unmute screenshare

What happens

1. Microphone control does not work as expected.
   - First device who joined can mute audio in both devices but only at the UI level for the second one. The first device's microphone turn off but the 2nd does not.
   - Second device cannot tap the microphone button at all

see screenshot below which shows both microphones are turned off in the first device but 2nd device shows the other device's microphone is turned on.

| 1st device | 2nd device |
|-----------|-------------|
|![Image](https://github.com/user-attachments/assets/c19fbb65-e3e3-4fed-9723-6b8254604c3f) | ![Image](https://github.com/user-attachments/assets/15595df6-9e18-456f-a334-44ec17455cc8) |

For the controls, I am using the provided `ToggleMicrophoneOption` in stream_video_flutter.

_Expected behavior: The device can control its associated participant._

2. Video control does not work as expected.

- First device can turn its video on and off except, it is showing in the UI as the video of the second device.
- Second device seems to show the same video for both users. It can also turn the video on and off on its UI.

See screenshots. First device is showing the video after its camera is turned on (which shows the camera input from 2nd device)

| 1st device | 2nd device |
|-----------|------------|
| ![Image](https://github.com/user-attachments/assets/09396fd1-837c-4e09-9120-2f024295c4c5) |  ![Image](https://github.com/user-attachments/assets/c6b8a7bc-c025-489f-93dd-a66b4138b77b)

_Expected behavior: The device can control its associated participant._

3. Screenshare does not work as expected

- First device can screenshare but second device does not display it
- Second device can also screenshare but the first device does not display it (although the expected behavior should be that this cannot screenshare when another user is screensharing

| 1st device | 2nd device |
|-----------|------------|
| ![Image](https://github.com/user-attachments/assets/e78e8a53-eafd-417b-bef4-b45cce64d2fb) | ![Image](https://github.com/user-attachments/assets/34fdd24f-f33f-4e55-8e0b-6d276e1ac22f) |

_Expected behavior: Each device is treated as a separate participant in the call._

---

I think these behaviors are to be expected given that the user is technically just the same participant which means they share the same `CallParticipantState`.

[...]

PR #915, opened 2025-03-21
fix(llc): fix checks for local participant

Brazol:
This pull request addresses a problem that occurs when a participant joins a call from multiple devices, leading to several issues with call controls. The fix ensures that we compare both the userId and sessionId, rather than just the userId, allowing us to differentiate between the same user participating in different sessions.

Fixes: https://github.com/GetStream/stream-video-flutter/issues/914
FLU-45

### DIFF
--- a/packages/stream_video/lib/src/call/call.dart
+++ b/packages/stream_video/lib/src/call/call.dart
@@ -2192,16 +2192,13 @@ class Call {
       );
 
       if (enabled) {
-        _session?.rtcManager
-            ?.getPublisherTrackByType(SfuTrackType.screenShare)
-            ?.mediaTrack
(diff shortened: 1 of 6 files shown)


tkortekaas:
Is this not going to cause the issue again where one user can connect twice using the same device? You can have two users from different sessions joining, but not two of the same user joining from the same device.

Brazol:
> Is this not going to cause the issue again where one user can connect twice using the same device? You can have two users from different sessions joining, but not two of the same user joining from the same device.

Hey @tkortekaas, to avoid duplicate users from the same device, we have checks at various stages in the `join` flow. This change only affects participants who have already joined, marking them as local or remote.

PR #918, opened 2025-03-26
fix(llc): fixes for fast reconnect flow

Brazol:
Fixes FLU-76

This pull request is focused on enhancing network monitoring and improving reconnection strategies. It addresses issues with the fast reconnect flow, specifically regarding the recreation of the SFU WebSocket, and includes various other fixes related to the reconnect process. Additionally, it unifies network connection monitoring across the SDK and provides more configuration options for better control.


packages/stream_video/lib/src/call/call.dart
```diff
@@ -683,17 +678,15 @@ class Call {
...
           _reconnectStrategy == SfuReconnectionStrategy.unspecified
               ? null
               : await _previousSession?.getReconnectDetails(_reconnectStrategy);
 
-      if (performingRejoin || performingMigration || !isWsHealthy) {
+      if (performingRejoin || performingMigration || !performingFastReconnect) {
         _logger.v(
           () =>
-              '[join] creating new sfu session (rejoin: $performingRejoin, migration: $performingMigration, wsHealthy: $isWsHealthy)',
+              '[join] creating new sfu session (rejoin: $performingRejoin, migration: $performingMigration, fast: $performingFastReconnect)',
```
renefloor, line 689: The original `_reconnectStrategy` is an enum, so isn't this the same as
```dart
if(!performingFastReconnect)
```
The logger also doesn't make sense if I understand it correctly, because `fast` is always `false`.

When `performingRejoin` or `performingMigration` are true `performingFastReconnect` has to be false.

packages/stream_video/lib/src/call/call.dart
```diff
@@ -743,21 +741,20 @@ class Call {
         }
       }
 
-      if (performingFastReconnect && isWsHealthy) {
+      if (performingFastReconnect) {
```
renefloor, line 744: If you simplify the previous `if`-statement, you can move this to the `else`

packages/stream_video/lib/src/call/session/call_session.dart
```diff
@@ -399,61 +404,65 @@ class CallSession extends Disposable {
...
+            (
+              callState: event.callState,
+              fastReconnectDeadline: event.fastReconnectDeadline
+            ),
+          );
+        } else {
+          _logger.v(() => '[fastReconnect] fast-reconnect not possible');
+          return const Result.failure(
+            VideoError(message: 'Fast reconnect not possible'),
+          );
```
renefloor, line 451: At some point we need to improve our exceptions so we can show better information to the users.

renefloor (approved):

PR #921, opened 2025-03-31
fix: call preference initialization during ringing

Brazol:
Fixes FLU-47

This pull request addresses the management of call preferences, particularly during the ringing flow. Previously, call preferences set while creating a ringing call could be overlooked if `StreamVideo.incomingCall` was set beforehand when handling the ringing event. Additionally, default call properties have been added to the StreamVideo options for improved functionality.


renefloor (approved):
This is a breaking change, but we are already making breaking changes in the next release.

PR #924, opened 2025-04-03
feat(llc): camera zoom and focus

Brazol:
Closes FLU-93

This PR adds camera focus and zoom control features to the SDK, allowing users to:
- Adjust the camera zoom level with Call.setZoom()
- Control focus and exposure points with Call.focus()

### DIFF
--- a/packages/stream_video/lib/src/call/call.dart
+++ b/packages/stream_video/lib/src/call/call.dart
@@ -1,6 +1,7 @@
 // ignore_for_file: deprecated_member_use_from_same_package
 
 import 'dart:async';
+import 'dart:math';
 import 'dart:typed_data';
 
 import 'package:collection/collection.dart';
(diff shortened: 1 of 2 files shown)


packages/stream_video/lib/src/call/call.dart
```diff
@@ -2137,6 +2138,62 @@ class Call {
     return const Result.success(false);
   }
 
+  Future<Result<None>> setZoom({
+    required double zoomLevel,
+  }) async {
```
renefloor, line 2143: I wonder if we should name this slightly different, like `setLocalCameraZoom`. Would be really nice if at some point we can zoom into other tracks, such as the screen sharing track. Apps like Microsoft teams allow for zooming on screen shares from other participants. I do see that the `rtc.helper` in the end really uses the camera utils on the native side, so doesn't make sense to create a generic `setZoom` for every track.

renefloor (approved):

PR #939, opened 2025-04-17
fix: call api fixes

Brazol:
This PR adds some missing parts of Call API. It also changes the way members are composed when creating a call.

### DIFF
--- a/packages/stream_video/lib/src/call/call.dart
+++ b/packages/stream_video/lib/src/call/call.dart
@@ -932,6 +932,7 @@ class Call {
 
   Future<Result<CallMetadata>> update({
     Map<String, Object>? custom,
+    DateTime? startsAt,
     StreamRingSettings? ring,
     StreamAudioSettings? audio,
     StreamVideoSettings? video,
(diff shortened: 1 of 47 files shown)


packages/stream_video/lib/src/call/call.dart
```diff
@@ -932,6 +932,7 @@ class Call {
 
   Future<Result<CallMetadata>> update({
     Map<String, Object>? custom,
+    DateTime? startsAt,
```
renefloor, line 935: Shouldn't we document this method?

packages/stream_video/lib/src/call/call.dart
```diff
@@ -1771,9 +1773,11 @@ class Call {
   /// - [notify]: If `true`, sends a standard push notification.
   /// - [video]: Marks the call as a video call if `true`; otherwise, audio-only.
   /// - [watch]:  If `true`, listens to coordinator events and updates call state accordingly.
-  /// - [membersLimit]: Sets the total number of members to return as part of the response.
```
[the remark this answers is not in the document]
Brazol: must have been by mistake, added again together with more docs

packages/stream_video/lib/src/call/call.dart
Brazol: added

renefloor (approved):

PR #954, opened 2025-05-13
Always create a new track on web

renefloor:
### 🎯 Goal

Fix screen sharing on web.
Fixes FLU-124

### 🛠 Implementation details

When starting to share your screen on web the `_setTrackEnabled` is called, but the deviceId is always `null` for the requirements. Therefore `compareScreenShareMode` is always true, even though we should create a new track.

The callback is then set to the wrong track, the old one: https://github.com/GetStream/stream-video-flutter/blob/main/packages/stream_video/lib/src/call/call.dart#L2428-L2430

Debugging showed that on the actual track the onEnded callback was `null`: https://github.com/flutter-webrtc/dart-webrtc/blob/main/lib/src/media_stream_track_impl.dart#L16

### 🧪 Testing

Test by running on web and start screen sharing only tabs and stop them using the browser tools.

#### General

### ☑️Reviewer Checklist
- [ ] Sample runs & works

### DIFF
--- a/packages/stream_video/lib/src/webrtc/rtc_track/rtc_local_track.dart
+++ b/packages/stream_video/lib/src/webrtc/rtc_track/rtc_local_track.dart
@@ -1,4 +1,5 @@
 import 'package:collection/collection.dart';
+import 'package:flutter/foundation.dart';
 import 'package:stream_webrtc_flutter/stream_webrtc_flutter.dart' as rtc;
 
 import '../../exceptions/video_exception.dart';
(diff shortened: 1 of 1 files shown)


packages/stream_video/lib/src/webrtc/rtc_track/rtc_local_track.dart
```diff
@@ -368,6 +369,11 @@ extension RtcLocalScreenShareTrackExt on RtcLocalScreenShareTrack {
       return true;
     }
 
+    // On web we always get a new media track and deviceId is `null`.
+    if (kIsWeb) {
```
xsahil03x, line 373: Consider using `CurrentPlatform.isWeb` instead.

Brazol (approved):
LGTM

PR #955, opened 2025-05-13
fix: fix app init delay

Brazol:
We had an artificial delay when initializing the app (to show the splash screen longer) that interfered with CallKit integration. When the app was started from a terminated state after a VoIP push was received (iOS), quickly accepting the call would not work because CallKit events were not observed.

### DIFF
--- a/dogfooding/lib/app/app.dart
+++ b/dogfooding/lib/app/app.dart
@@ -31,9 +31,7 @@ class _StreamDogFoodingAppState extends State<StreamDogFoodingApp> {
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
-    _appLoader ??= Future.wait([
-      _loadApp(context),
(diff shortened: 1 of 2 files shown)


dogfooding/lib/app/app.dart
```diff
@@ -33,8 +33,6 @@ class _StreamDogFoodingAppState extends State<StreamDogFoodingApp> {
     super.didChangeDependencies();
     _appLoader ??= Future.wait([
       _loadApp(context),
-      // Shows the splash screen for at least 3 seconds.
-      Future.delayed(const Duration(seconds: 3)),
     ]);
```
renefloor, line 36: We only have 1 future left, so I would skip the list.

```dart
    _appLoader ??= _loadApp(context);
```

renefloor (approved):

PR #957, opened 2025-05-14
Store previous sorting of participants before sorting again

renefloor:
### 🎯 Goal

Fixes FLU-82
Prevent jumping around of participant locations

### 🛠 Implementation details

Some sorting rules are not applied when participants are visible. However, because we weren't storing the result of the previous sort participants were jumping back to their previous position. By keeping the sort result and applying that to the new set of participants we make sure participants only jump around when we want them to.

### 🧪 Testing

You can start a call with many participants and try to play with who's talking or screen sharing for example.

#### General

### ☑️Reviewer Checklist
- [ ] Sample runs & works


Brazol (approved):
LGTM

PR #959, opened 2025-05-15
chore: call duration stream

Brazol:
resolves FLU-136

### 🎯 Goal

We aim to simplify the process for integrators to obtain the current call duration by providing them with a stream of the duration in seconds.

### 🛠 Implementation details

Added a Stream<int> in `CallStateNotifier` that is started whenever `startedAt` or `liveStartedAt` is set. The stream is exposed in `Call` as `callDurationSecondsStream`. 

### 🎨 UI Changes

The new stream is now also used in Dogfooding timer widget and `LivestreamPlayer` widget.

### DIFF
--- a/packages/stream_video/lib/src/call/state/call_state_notifier.dart
+++ b/packages/stream_video/lib/src/call/state/call_state_notifier.dart
@@ -25,7 +25,7 @@ class CallStateNotifier extends StateNotifier<CallState>
     callStateStream =
         MutableStateEmitterImpl<CallState>(initialState, sync: true);
     _durationTimerController = StreamController<Duration>.broadcast();
-    _ensureStartDurationTimer();
+    _setupDurationTimer();
   }
 
   final _logger = taggedLogger(tag: 'CallStateNotifier');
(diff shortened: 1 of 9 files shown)


packages/stream_video/lib/src/call/state/call_state_notifier.dart
```diff
@@ -22,13 +24,19 @@ class CallStateNotifier extends StateNotifier<CallState>
...
+    _durationTimerController = StreamController<int>.broadcast();
+    _ensureStartDurationTimer();
   }
 
   final _logger = taggedLogger(tag: 'CallStateNotifier');
 
   late final MutableStateEmitterImpl<CallState> callStateStream;
   CallState get callState => callStateStream.value;
 
+  Stream<int> get durationStream => _durationTimerController.stream.distinct();
```
renefloor, line 36: Even though we update once per second I think I would still do the duration as Duration object.

packages/stream_video/lib/src/call_state.dart
```diff
@@ -19,7 +19,7 @@ class CallState extends Equatable {
       preferences: preferences,
       currentUserId: currentUserId,
       callCid: callCid,
-      createdByUserId: '',
```
renefloor, line 22: Did you mix this PR with another fix by accident?

Brazol: I also replaced the `createdByUserId` with `createdByUser` in this PR to solve two gaps that Martin reported we had in CallState.

renefloor (approved):

PR #964, opened 2025-05-21
fix: do not unsubscribe screen share track when zooming

Brazol:
resolves FLU-131

This PR addresses an issue with our track visibility logic. We currently unsubscribe from video tracks when their visibility falls below 30% to conserve resources. However, this had an unintended side effect: when users zoom in on a screen share track using InteractiveViewer, the zoom could reduce the track's visible area below the threshold, causing it to be unsubscribed.

This fix ensures that the screen share track remains subscribed even when zoomed in, preserving the user experience during screen sharing.

### DIFF
--- a/packages/stream_video_flutter/lib/src/call_participants/screen_share_call_participants_content.dart
+++ b/packages/stream_video_flutter/lib/src/call_participants/screen_share_call_participants_content.dart
@@ -131,7 +131,7 @@ class ScreenShareContent extends StatefulWidget {
 }
 
 class _ScreenShareContentState extends State<ScreenShareContent> {
-  double _currentScale = 1;
+  bool _isZoomed = false;
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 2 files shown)


packages/stream_video_flutter/lib/src/call_participants/screen_share_call_participants_content.dart
```diff
@@ -146,11 +153,18 @@ class ScreenShareContent extends StatelessWidget {
...
+                });
+              },
               child: StreamVideoRenderer(
-                call: call,
-                participant: participant,
+                call: widget.call,
+                participant: widget.participant,
                 videoFit: VideoFit.contain,
                 videoTrackType: SfuTrackType.screenShare,
+                persistTrackIfNotVisible: _currentScale > 1.0,
```
renefloor, line 167: I wonder if this adds way to many rebuilds. Because we're setting the state every time you zoom a bit.

Can't we just always set this to true always, or only update the state if the zoom starts/stops? For example:
```dart
              onInteractionUpdate: (details) {
                final isZoomed = details.scale > 1.0;
                if(_isZoomed != isZoomed){
                  setState(() {
                    _isZoomed = isZoomed;
                  });
                },
              },
```

renefloor (approved):

PR #970, opened 2025-05-26
chore: default audio logic in line with other sdks

Brazol:
part of FLU-147
resolves FLU-148

### 🎯 Goal

Change the default audio output logic based on Dashboard configuration to match what other SDKs are doing. We want to proritise external device and then set speaker if video camera is on, speaker setting is enabled, or explicitly set as default device.

### 🛠 Implementation details

Changed the logic in `_applyCallSettingsToConnectOptions` to match what we wanted to achieve.
Added `isExternal` getter to `RtcMediaDevice` to determine if external audio device is connected.

### 🎨 UI Changes

Changed audio output picker in Dogfooding app to show native iOS picker on iOS devices.


packages/stream_video/lib/src/call/call.dart
```diff
@@ -1494,15 +1525,11 @@ class Call {
           .contains(settings.video.cameraFacing.value.toLowerCase()),
     );
 
+    // If it's not front or back then take one of the external cameras
     if (defaultVideoInput == null && videoInputs.length > 2) {
-      // If it's not front or back then take one of the external cameras
       defaultVideoInput = videoInputs.last;
```
renefloor, line 1530: Why is this the last?

packages/stream_video/lib/src/call/call.dart
```diff
@@ -1471,21 +1470,53 @@ class Call {
...
+        );
+      } else {
+        // 3. First non-speaker device
+        defaultAudioOutput = audioOutputs.firstWhereOrNull(
+          (device) => !device.id.equalsIgnoreCase(
+            AudioSettingsRequestDefaultDeviceEnum.speaker.value,
+          ),
         );
       }
+    }
```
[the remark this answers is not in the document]
Brazol: Hm actually I think we should set null for `defaultAudioInput` instead of `audioInputs.firstOrNull`. This way we leave it for the OS to decide and handle. From my experience while testing this the less we intrude the better. So we should force the device only if we really need to.

packages/stream_video/lib/src/call/call.dart
Brazol: not sure, looks wrong. Lets address the video input config in separate PR/task.

renefloor (approved):

renefloor (approved):

PR #974, opened 2025-05-29
fix(llc): fix update permission event handling

Brazol:
### 🎯 Goal

Fix for updating permission event handling. OwnCapabilities should be updated only if the event targets the current user. Additionally, handling of member permission update events has been added.

### DIFF
--- a/packages/stream_video/lib/src/call/state/mixins/state_coordinator_mixin.dart
+++ b/packages/stream_video/lib/src/call/state/mixins/state_coordinator_mixin.dart
@@ -163,6 +163,13 @@ mixin StateCoordinatorMixin on StateNotifier<CallState> {
   void coordinatorCallPermissionsUpdated(
     StreamCallPermissionsUpdatedEvent event,
   ) {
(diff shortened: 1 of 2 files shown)


packages/stream_video/lib/src/call/state/mixins/state_coordinator_mixin.dart
```diff
@@ -498,6 +506,7 @@ mixin StateCoordinatorMixin on StateNotifier<CallState> {
           return member;
         }
       }).toList(),
+      capabilitiesByRole: capabilitiesByRole,
```
renefloor, line 509: In `coordinatorCallPermissionsUpdated` you check for the current user id. Are we certain that `StreamCallMemberUpdatedPermissionEvent` can only be for the current user?

Brazol: This event updates the `members` list and `capabilitiesByRole` map, and those are global for the call. It does not update anything specific to the single user.

PR #977, opened 2025-06-02
feat: allow multiple simultaneous active calls

Brazol:
resolves FLU-158

### 🎯 Goal

Right now, our Video SDK only allows one active call at a time. We keep track of the active call in a state, and when a new call connects, we always end the previous one. We want to allow multiple active calls to happen in the same time as some clients have this requirement. This PR makes it possible.

### 🛠 Implementation details

- By default, only a single active call is still permitted, as it was before. To change this behavior, `allowMultipleActiveCalls` has been added to `StreamVideoOptions`. 
- All places that relied on activeCall are adjusted to handle multiple active calls.
- `StreamBackgroundService` is also adjusted, showing multiple notifications


packages/stream_video_flutter/lib/stream_video_flutter_background.dart
```diff
@@ -41,18 +41,20 @@ mixin StreamVideoFlutterBackground {
     );
   }
 
-  static Future<bool> stopService(ServiceType type) async {
```
renefloor, line 44: Is there an option to make this `callId` optional and check if there is only 1 call running? Would prevent a breaking change and makes code slightly easier for when you don't allow for multiple calls.

Or maybe to make it more clear we could add a different method like this:

```dart
static Future<bool> stopService(ServiceType type) {
  final callCid = getCurrentCall();
  return stopServiceForCall(type, callCid: callCid);
}

static Future<bool> stopServiceForCall(ServiceType type, {required String callCid}) async {
...
}

File: packages/stream_video/lib/src/core/client_state.dart [hunk 2]
```diff
@@ -17,7 +20,7 @@ abstract class ClientState {
   StateEmitter<ConnectionState> get connection;
 
   /// Sets when a call being joined.
-  StateEmitter<Call?> get activeCall;
+  StateEmitter<List<Call>> get activeCalls;
```
renefloor, line 23: To prevent breaking changes we could keep the old one as well, but deprecate it.

I don't think this is the right place for this logic, but something similar like this:

I'm also not sure if we even need to deprecate it. Maybe just a nice feature anyway for apps that never have multiple active calls.

```dart
  /// Returns first of active calls or null if no calls are active
  @Deprecated("please use `activeCalls`")
  StateEmitter<Call?> get activeCall => activeCalls.firstOrNull;
  
  /// List of currently active calls.
  StateEmitter<List<Call>> get activeCalls;
```

packages/stream_video/lib/src/core/client_state.dart
```diff
@@ -58,22 +65,32 @@ class MutableClientState implements ClientState {
...
 
   Future<void> clear() async {
-    activeCall.value = null;
+    activeCalls.value = [];
     outgoingCall.value = null;
     connection.value = ConnectionState.disconnected(user.value.id);
   }
 
-  Call? getActiveCall() => activeCall.valueOrNull;
+  List<Call>? getActiveCalls() => activeCalls.valueOrNull;
```
renefloor, line 73: I think this doesn't have to be nullable?

packages/stream_video/lib/src/core/client_state.dart
```diff
@@ -26,24 +29,28 @@ abstract class ClientState {
   StateEmitter<Call?> get outgoingCall;
 
   Future<void> setOutgoingCall(Call? call);
-  Future<void> setActiveCall(Call? call);
+
+  Future<void> setActiveCall(Call call);
```
renefloor, line 33: If we want to prevent breaking changes we could allow setting `null` here still, and just cancel all active calls in that case.

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -245,8 +246,7 @@ class StreamVideo extends Disposable {
   final StreamVideoOptions _options;
   final MutableClientState _state;
 
-  bool get muteVideoWhenInBackground => _options.muteVideoWhenInBackground;
-  bool get muteAudioWhenInBackground => _options.muteAudioWhenInBackground;
```
renefloor, line 249: I would keep them, but deprecate them.

renefloor (commented):
I added some comments about backwards compatibility. 

`client_state.dart` is not exported from the library, but is accessible through `StreamVideo.state`, so it's technically part of the public API.

Not sure if we can completely prevent breaking changes, but I would really prefer to start limiting them whenever possible. Especially in this case, where preventing the breaking changes also makes it easier for most developers, as having multiple active calls isn't the main usecase of the libary.

packages/stream_video/lib/src/core/client_state.dart
```diff
@@ -58,22 +73,54 @@ class MutableClientState implements ClientState {
...
+    }
+
+    if (call == null) {
+      activeCalls.value = [];
+    } else if (!options.allowMultipleActiveCalls) {
+      activeCalls.value = [call];
+    } else {
+      activeCalls.value = [...activeCalls.value, call];
+    }
+  }
```
renefloor, line 112: I think we should only do this in case 1 call is allowed:
```dart
    if (call == null) {
      activeCalls.value = [];
    }
```
So I would move that up to the `if (!options.allowMultipleActiveCalls)` and merge it together. You have 3 cases now only for `if (!options.allowMultipleActiveCalls)`.

```suggestion
    if (!options.allowMultipleActiveCalls) {
      final currentlyActiveCall = activeCalls.value.firstOrNull;
      if (currentlyActiveCall != null) {
        await currentlyActiveCall.leave(reason: DisconnectReason.replaced());
      }

      activeCall.value = call;
      activeCalls.value = call == null ? [] : [call];
    } else if (call != null) {
      activeCalls.value = [...activeCalls.value, call];
    }
  }
```

packages/stream_video_flutter/lib/src/call_background/background_service.dart
```diff
@@ -45,237 +58,269 @@ class StreamBackgroundService {
...
-  return const NotificationOptions(
+  final options = NotificationOptions(
     content: NotificationContent(
-      title: 'You are screen sharing',
+      title: 'You are screen sharing (${call.callCid.id})',
+      text: 'Tap to return to the call.',
     ),
   );
+
+  return options;
```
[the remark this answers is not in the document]
Brazol: just a leftover after debugging, fixed

PR #1024, opened 2025-07-15
feat(llc): Make sure the audio input device is set even when the user joins muted

Brazol:
resolves FLU-202

When the user joins the call muted, the audio track is not created. It will be created the first time he or she unmutes. If the user wants to set the audio input device during that time, it would fail. We want to make it possible by storing the chosen deviceId and applying the change when the track is created.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where changing the audio input device while muted at the start of a call did not apply the new device upon unmuting; the selected device is now correctly set when unmuted.
  * Improved error handling for missing audio tracks, ensuring smoother device switching and unmuting behavior.
  * Minor update to changelog formatting for clarity.

### DIFF
--- a/packages/stream_video/test/src/call/call_test.dart
+++ b/packages/stream_video/test/src/call/call_test.dart
@@ -15,13 +15,6 @@ void main() {
   setUpAll(() {
     registerMockFallbackValues();
     TestWidgetsFlutterBinding.ensureInitialized();
-
-    registerFallbackValue(const RtcMediaDevice(
-      id: 'fallback-device',
-      label: 'Fallback Device',
-      kind: RtcMediaDeviceKind.audioInput,
-      groupId: 'fallback-group',
-    ));
   });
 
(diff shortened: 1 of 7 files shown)


packages/stream_video/test/src/call/call_test.dart
```diff
@@ -13,6 +15,13 @@ void main() {
   setUpAll(() {
     registerMockFallbackValues();
     TestWidgetsFlutterBinding.ensureInitialized();
+
+    registerFallbackValue(const RtcMediaDevice(
+      id: 'fallback-device',
+      label: 'Fallback Device',
+      kind: RtcMediaDeviceKind.audioInput,
+      groupId: 'fallback-group',
+    ));
```
renefloor, line 24: You can add this directly in the `registerMockFallbackValues` method.

PR #1032, opened 2025-07-22
Update screen share visibility separate from video

renefloor:
### 🎯 Goal

Fixes FLU-210

### 🛠 Implementation details

Participants were jumping, because participant visibility was set to hidden when the screen share track was hidden.

### 🧪 Testing

Easily test before/after by having a call with 2 members and share screen from one of them. When you stop sharing the participant who talks is always put on spot 1 even though both are always visible. After the fix it works normally.

#### General

### ☑️Reviewer Checklist
- [ ] Sample runs & works

### DIFF
--- a/packages/stream_video/lib/src/call/call.dart
+++ b/packages/stream_video/lib/src/call/call.dart
@@ -2772,12 +2772,21 @@ class Call {
     required String sessionId,
     required String userId,
     required ViewportVisibility visibility,
+    required SfuTrackTypeVideo trackType,
   }) async {
     final change = VisibilityChange(
       sessionId: sessionId,
       userId: userId,
       visibility: visibility,
     );
+    if (trackType.isScreenShare) {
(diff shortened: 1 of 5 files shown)


packages/stream_video/lib/src/call/call.dart
```diff
@@ -2772,12 +2772,21 @@ class Call {
...
+      _stateManager.participantUpdateScreenShareViewportVisibility(
+        sessionId: sessionId,
+        userId: userId,
+        visibility: visibility,
+      );
+      return const Result.success(none);
+    }
 
     final result = await _session?.updateViewportVisibility(change) ??
         Result.error('Session is null');
```
renefloor, line 2792: Not sure if we should do anything with the session here.

Brazol: I think it's fine to have it in the participant state.

Brazol (approved):

PR #1044, opened 2025-08-07
feat: migrated callkit incoming into stream_video_push_notification

Brazol:


packages/stream_video_push_notification/lib/src/stream_video_push_configuration.dart
```diff
@@ -0,0 +1,283 @@
+import 'package:json_annotation/json_annotation.dart';
+import 'package:stream_video_push_notification/stream_video_push_notification.dart';
+
+part 'stream_video_push_configuration.g.dart';
+
+@JsonSerializable(explicitToJson: true)
```
renefloor, line 6: I think we can also add `explicitToJson` in the `build.yaml`, or are there also cases where we don't want that?

PR #1058, opened 2025-09-01
feat(llc): kick user feature

Brazol:
resolves FLU-238
docs: https://github.com/GetStream/docs-content/pull/580

Added option to remove participant from the call by calling `call.kickUser()` method.

## Summary by CodeRabbit

* **New Features**
  * Kick participants from a call; configure ingress (incoming) audio/video encoding and layers when creating/updating calls.

* **Events**
  * New "user kicked" event emitted, including who performed the action.

* **Permissions**
  * Added "kick-user" permission for participant removal.

* **User Data**
  * avgResponseTime exposed on user profiles/responses/privacy fields.

* **Analytics**
  * Video events include rating, SDK name/version, and kicked-by user.

* **Documentation**
  * Changelog updated to document participant removal via call.kickUser().

### DIFF
--- a/dogfooding/ios/fastlane/Fastfile
+++ b/dogfooding/ios/fastlane/Fastfile
@@ -78,8 +78,8 @@ platform :ios do
         groups: is_prod ? ['Internal Testers', 'External Testers'] : ['Internal Testers'],
         changelog: 'Lots of amazing new features to test out!',
         reject_build_waiting_for_review: false,
-        skip_waiting_for_build_processing: false,
-        beta_app_description: 'Beta build of Stream Video Flutter Dogfoofing app for internal and external testers',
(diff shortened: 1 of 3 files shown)


packages/stream_video/lib/src/call/call_events.dart
```diff
@@ -1037,6 +1037,28 @@ class StreamCallUserUnblockedEvent extends StreamCallEvent {
...
+
+  @override
+  List<Object?> get props => [
+        ...super.props,
+        createdAt,
+        user,
+        kickedByUser,
+      ];
+}
+
```
renefloor, line 1061: Should we also leave the call when we are kicked? Or do we get another call event that would trigger a leave?

PR #1063, opened 2025-09-15
feat(llc): video pause feature

Brazol:
This PR adds automatic SFU-driven pausing of inbound video tracks to save bandwidth and prevent visual artifacts on constrained networks.

### DIFF
--- a/packages/stream_video/lib/src/call/call.dart
+++ b/packages/stream_video/lib/src/call/call.dart
@@ -29,6 +29,7 @@ import '../models/models.dart';
 import '../platform_detector/platform_detector.dart';
 import '../retry/retry_policy.dart';
 import '../sfu/data/events/sfu_events.dart';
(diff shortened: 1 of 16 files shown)


packages/stream_video/lib/src/call/call.dart
```diff
@@ -550,6 +554,20 @@ class Call {
...
+    capabilities.forEach(_sfuClientCapabilities.add);
+  }
+
+  void disableClientCapabilities(
+    List<SfuClientCapability> capabilities,
+  ) {
+    _logger.i(() => '[disableClientCapabilities] $capabilities');
+    capabilities.forEach(_sfuClientCapabilities.remove);
+  }
+
```
renefloor, line 570: I think this is public API right? If I understand correctly you can only do this before you join the call? We should document that.

renefloor (approved):

Issue #1083, opened 2025-10-14
On iOS after leaving call and putting app to background the PiP mode is still on.


asaarnak:
I awaited `call.leave()` and after that navigated to previous screen so all the `StreamCallContainer` was disposed, then if i put my app to background in a second or two the PiP window with black screen is opened.

Brazol:
Hi @asaarnak, the issue should be resolved in the next version that we plan to release very soon.

PR #1101, opened 2025-11-06
fix(ui): fixes for PiP mode on iOS

Brazol:
resolves FLU-317, FLU-309, https://github.com/GetStream/stream-video-flutter/issues/1083

This PR fixes multiple issues in iOS Picture in Picture mode:
* Resolved an issue in Picture in Picture where video tracks might remain disabled after returning the app to the foreground.
* Addressed a problem where Picture in Picture was not exited properly if the call ended during PiP mode.
* Fixed a bug where quickly backgrounding the app right after ending a call could still activate PiP mode.

## Bug Fixes

* **iOS Picture-in-Picture** - Fixed video tracks remaining disabled after returning app to foreground from PiP mode.
* **iOS Picture-in-Picture** - Fixed improper PiP exit when a call ends during PiP mode.
* **iOS Picture-in-Picture** - Fixed unintended PiP activation when backgrounding the app immediately after call termination.


packages/stream_video_flutter/lib/src/call_screen/call_content/picture_in_picture/stream_picture_in_picture_ui_kit_view.dart
```diff
@@ -161,16 +177,59 @@ class _StreamPictureInPictureUiKitViewState
...
         break;
       default:
         break;
     }
   }
 
+  // When app goes to background disable video tracks of all participants except
+  // the one shown in PiP to save bandwidth and resources.
+  // Re-enable them when app comes back to foreground.
+  void _toggleSubscribedTracks(bool enable) {
```
xsahil03x: Will this also automatically update if the speaker changes while in background?

xsahil03x (approved):

PR #1241, opened 2026-06-01
chore(llc): noise cancellation enum rename

Brazol:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected noise-cancellation mode naming to ensure auto-on behavior works reliably during call start and teardown.

* **Tests**
  * Updated audio-processing tests to use the corrected noise-cancellation mode values.

* **Documentation**
  * Clarified CHANGELOG entry to reflect the accurate noise-cancellation mode names and related behavior.

### DIFF
--- a/packages/stream_video/lib/src/models/call_settings.dart
+++ b/packages/stream_video/lib/src/models/call_settings.dart
@@ -1296,6 +1296,7 @@ enum RecordingType {
   }
 }
 
+@Deprecated('Use NoiseCancellationSettingsMode instead.')
 typedef NoiceCancellationSettingsMode = NoiseCancellationSettingsMode;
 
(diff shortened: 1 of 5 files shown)


packages/stream_video/lib/src/models/call_settings.dart
```diff
@@ -1296,9 +1296,9 @@ enum RecordingType {
   }
 }
 
-typedef NoiseCancellationSettingsMode = NoiceCancellationSettingsMode;
+typedef NoiceCancellationSettingsMode = NoiseCancellationSettingsMode;
```
xsahil03x, line 1300: let's also mark it deprecated

xsahil03x (approved):

PR #1261, opened 2026-06-22
feat(llc): new openAPI generator + stream core dependency

Brazol:
Open API models generated using generator from chat repository using `generate.sh` script.
Added dependency to stream_core and replaced SDKs implementations of `Result`, SharedEmitter`, `StateEmitter`, `CurrentPlatform`, `Logger` with implementation from core.


packages/stream_video/lib/open_api/video/coordinator/model/attachment.dart
```diff
@@ -0,0 +1,116 @@
...
+@JsonSerializable()
+class Attachment with _$Attachment {
+  const Attachment({
+    this.actions,
+    this.assetUrl,
+    this.authorIcon,
+    this.authorLink,
+    this.authorName,
+    this.color,
+    required this.custom,
```
xsahil03x: I was wondering for cases like these, if we have a default value set in json, should we also set it in the constructor and mark it non-required?

packages/stream_video/lib/src/coordinator/open_api/coordinator_client_open_api.dart
```diff
@@ -61,21 +49,21 @@ class CoordinatorClientOpenApi extends CoordinatorClient {
...
+          final tokenResult = await _tokenManager.getToken();
+          if (tokenResult is! Success<UserToken>) {
+            throw (tokenResult as Failure).videoError;
+          }
+          return tokenResult.data;
+        },
+        getConnectionId: () => _ws?.connectionId,
+      ),
+    );
+  late final _defaultApi = open.DefaultApi(_dio);
```
xsahil03x: Do you think we can use the `StreamCoreHttpClient` here?

```dart
final httpClient =
        StreamCoreHttpClient(
          options: BaseOptions(
            baseUrl: endpointConfig.baseFeedsUrl,
            connectTimeout: const Duration(seconds: 6),
            receiveTimeout: const Duration(seconds: 6),
          ),
        ).apply(
          (client) => client.interceptors.addAll([
            ApiKeyInterceptor(apiKey),
            HeadersInterceptor(_systemEnvironmentManager),
            if (user.type != UserType.anonymous) connectionIdInterceptor,
            AuthInterceptor(client, _tokenManager),
            const ApiErrorInterceptor(),
            LoggingInterceptor(requestHeader: true),
          ]),
        );
```

packages/stream_video/lib/src/coordinator/open_api/coordinator_client_open_api.dart
```diff
@@ -1601,7 +1751,7 @@ class CoordinatorClientOpenApi extends CoordinatorClient {
 typedef GetConnectionId = String? Function();
 typedef GetToken = FutureOr<UserToken> Function();
 
-class _Authentication extends open.Authentication {
+class _Authentication extends Interceptor {
```
xsahil03x: Instead of a single interceptor, can we use separate ones similar to feeds? the interceptors are already available in core.

https://github.com/GetStream/stream-feeds-flutter/blob/main/packages/stream_feeds/lib/src/client/feeds_client_impl.dart#L145-L150

xsahil03x: Maybe good to add all the interceptors like we have in feeds

packages/stream_video/lib/src/coordinator/open_api/open_api_extensions.dart
```diff
@@ -507,9 +505,12 @@ extension on open.IngressVideoLayerResponse {
...
     return StreamIngressVideoLayer(
       bitrate: bitrate,
-      codec:
-          IngressVideoLayerRequestCodecEnumTypeTransformer().decode(codec) ??
-          IngressVideoLayerRequestCodecEnum.h264,
+      codec: switch (codec) {
+        'h264' => IngressVideoLayerRequestCodec.h264,
+        'vp8' => IngressVideoLayerRequestCodec.vp8,
+        '_unknown' => IngressVideoLayerRequestCodec.unknown,
+        _ => IngressVideoLayerRequestCodec.h264,
```
xsahil03x: Maybe better to create an extension type for the raw strings

packages/stream_video/lib/src/platform_detector/platform_detector_io.dart
xsahil03x, line 1: We can remove all the platform detector files.

packages/stream_video/lib/src/utils/result.dart
```diff
@@ -1,210 +1,87 @@
...
-    switch (_type) {
-      case _ResultType.success:
-        return success?.call((this as Success<T>).data);
-      case _ResultType.failure:
-        return failure?.call((this as Failure).error);
-    }
-  }
+/// Backwards-compatible accessors that existed on the previous `stream_video`
+/// [Result] but are not part of the `stream_core` one.
+/// Added to minimize the impact of the breaking changes.
```
xsahil03x, line 23: We should mark the whole extension deprecated so that we know we also have to migrate to the new ones later.

packages/stream_video/CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
...
+  - Equal consecutive values are **conflated** — setting the same value twice emits only once. Previously every assignment produced an event regardless of equality.
+  - `StateEmitter` and `SharedEmitter` now **implement `Stream` directly**. Use the emitter itself (e.g. `.listen()`, `.firstWhere()`) instead of `.asStream()` or `.valueStream`.
+  - `MutableStateEmitterImpl<T>` is now a typedef for `StateEmitterImpl<T>` from `stream_core`. Existing call sites continue to compile, but the concrete type has changed.
+  - `MutableSharedEmitterImpl<T>` is now a typedef for `MutableSharedEmitter<T>` from `stream_core`. Existing call sites continue to compile, but the concrete type has changed.
+  - `SharedEmitter.firstWhere` no longer accepts a `timeLimit` parameter — use `waitFor` with an optional `timeLimit`, or call `.firstWhere(...).timeout(...)`.
+- `CurrentPlatform` and `PlatformType` are now provided by `stream_core`:
+  - `CurrentPlatform.name` has been removed — use `CurrentPlatform.operatingSystem` instead.
+- `StreamLogger`, `Priority`, and `MessageBuilder` are now provided by `stream_core`. Import them from `stream_video` as before — the re-export is in place.
+
+## 1.4.1
```
xsahil03x, line 23: Is the branch not up to date with master? We are seeing the master changes here.

packages/stream_video/lib/open_api/video/coordinator/model/attachment.dart
Brazol: Makes sense. I added this to the generators: https://github.com/GetStream/chat/pull/14119/changes/724b8e3887c55d9cab8ee606f0f53f4018addc22
And regenerated the models

packages/stream_video/lib/src/coordinator/open_api/coordinator_client_open_api.dart
Brazol: yes, missed the fact that it exists

packages/stream_video/lib/src/coordinator/open_api/coordinator_client_open_api.dart
Brazol: I'll reuse what we can from core and split the existing into smaller interceptors

packages/stream_video/lib/src/utils/result.dart
Brazol: the idea was to keep them. If I mark it deprecated I could just remove them now since we are doing a major release. I think they are harmless and quite useful

packages/stream_video/CHANGELOG.md
Brazol: updated `v2` with `master`

xsahil03x (approved):

PR #1262, opened 2026-06-24
fix(llc): fix stream header client details

Brazol:
## Summary by CodeRabbit

* **New Features**
  * Improved device and app environment reporting for the SDK, including browser and WebRTC details where available.
  * Added a more consistent client header generated from the current runtime environment.

* **Bug Fixes**
  * Made startup environment collection more reliable before connecting.
  * Reduced the risk of missing or outdated client details by centralizing environment updates.

### DIFF
--- a/packages/stream_video/lib/src/stream_video.dart
+++ b/packages/stream_video/lib/src/stream_video.dart
@@ -13,7 +13,6 @@ import 'package:system_info2/system_info2.dart';
 import 'package:uuid/uuid.dart';
 
 import '../globals.dart';
-import 'video_environment.dart';
 import '../open_api/video/coordinator/api.dart';
 import 'audio_processing/audio_processor.dart';
 import 'call/call.dart';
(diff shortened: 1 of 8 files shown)


xsahil03x (approved):

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -235,30 +235,26 @@ class StreamVideo extends Disposable {
     _setupLogger(options.logPriority, options.logHandlerFunction);
 
     unawaited(
-      _setClientDetails().onError((dynamic error, StackTrace stackTrace) {
-        _logger.e(
-          () =>
-              '[StreamVideo] failed to set client details: $error with stackTrace: $stackTrace',
-        );
+      _collectEnvironment().then((env) {
```
xsahil03x: not sure, but can you make sure we get the `.then` callback and calls connect even if the `_collectEnvironment` throws?

xsahil03x: maybe better to add a test for this

Brazol: yea nice catch, fixed and added tests

PR #1271, opened 2026-07-07
fix(ui): make video fit adaptive on web and desktop

Brazol:
## Summary by CodeRabbit

* **New Features**
  * Video rendering now adapts more intelligently to the current stream orientation and aspect ratio, improving how video fills the screen.
  * Default video fit behavior now automatically matches the platform, with different defaults on mobile versus web/desktop.

* **Bug Fixes**
  * Video display settings now fall back more consistently when no explicit fit mode is chosen.
  * Participant video appearance settings now support platform-based defaults for a more consistent viewing experience.


packages/stream_video_flutter/lib/src/renderer/video_renderer.dart
```diff
@@ -247,6 +250,21 @@ enum VideoFit {
   /// Scale the video uniformly (maintain the video's aspect ratio)
   /// to cover the entire widget area.
   cover,
+
+  /// Automatically pick between [cover] and [contain] based on the video's
+  /// own orientation: landscape (or square) video fills the widget ([cover]),
+  /// while portrait/tall video is letterboxed so the whole frame stays
+  /// visible ([contain]).
+  adaptive,
```
renefloor, line 258: I think on web they use `fillHeight`, because if the video is landscape, but the frame is wider than the video I still get black boxes on the side. Is that something we can add?

Brazol: adaptive fit will choose `cover` for landscape so it shouldn't leave black boxes on the sides. Anyway the `rtc.RTCVideoViewObjectFit` offers only `cover` or `contain` so we can't do anything else without additional work on webrtc side.

renefloor (approved):

PR #1275, opened 2026-07-08
fix(llc): fixed stats not being flushed on call end

Brazol:
## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured SFU stats are flushed and delivered when a call ends or reconnects, before teardown continues.
  * Improved call leave tracing to record the provided leave reason accurately.
* **Tests**
  * Added coverage for SFU stats reporter `flush()` to confirm stats are sent when applicable and not sent after shutdown.


packages/stream_video/lib/src/call/stats/sfu_stats_reporter.dart
```diff
@@ -116,22 +119,42 @@ class SfuStatsReporter {
...
+      _logger.v(() => 'Failed to flush SFU stats: $e');
+    }
+  }
+
   Future<void> sendSfuStats({
     int? connectionTimeMs,
     SfuReconnectionStrategy? reconnectionStrategy,
   }) async {
-    if (_stopped) return;
+    if (_stopped && !_flushing) return;
```
renefloor: Maybe easier to create a separate getter like
```
bool get _canSendSfuStats => _stopped && !_flushing
```

Brazol: done

renefloor: I guess it's my mistake, but the boolean is now inverted right?
```dart
bool get _canSendSfuStats => _stopped && !_flushing;
```

Everywhere you do 
```dart
if (_canSendSfuStats) return;
```
But I would expect `if (!_canSendSfuStats) return;`

renefloor (approved):

PR #1287, opened 2026-07-22
chore(llc): improve ice candidate tracing

Brazol:
## Summary by CodeRabbit

* **New Features**
  * ICE candidate handling now reports whether candidates are applied immediately or buffered until a remote description is available.
  * Connection tracing now distinguishes pending, successful, and failed ICE candidate addition.

* **Bug Fixes**
  * Prevent ICE candidates from being dropped during mid-negotiation by synchronizing remote-description updates and buffered candidate flushing.
  * Improved ICE candidate result handling across publisher/subscriber flows.
  * Republish now reuses a cached publisher transceiver without triggering unnecessary renegotiation.
  * Sanitized non-ASCII app/device info to avoid `X-Stream-Client` header `FormatException`.

* **Tests**
  * Expanded renegotiation/ICE buffering test coverage, including mid-operation and partial-flush scenarios.


packages/stream_video/lib/src/webrtc/peer_connection.dart
```diff
@@ -264,17 +277,21 @@ class StreamPeerConnection extends Disposable {
...
+        // Flush buffered candidates.
+        final pending = List<rtc.RTCIceCandidate>.of(_pendingCandidates);
+        for (final candidate in pending) {
+          await pc.addCandidate(candidate);
+          _pendingCandidates.remove(candidate);
+        }
+        return Result.success(result);
+      } catch (e, stk) {
+        return Result.failure(VideoErrors.compose(e, stk));
       }
```
renefloor, line 289: One bad candidate now poisons every subsequent setRemoteDescription (medium)

packages/stream_video/lib/src/webrtc/peer_connection.dart â the flush is inside the same try as the SDP set, so an addCandidate throw makes setRemoteDescription return Failure even though the remote description was applied successfully. Combined with the new retention-on-failure, a permanently-invalid candidate (malformed SDP mid, stale ufrag after an ICE restart) stays buffered and re-fails every future setRemoteDescription on that connection. Your own second test demonstrates the mechanism â it just happens to clear failOnCandidate before the retry.

Concretely: _onRenegotiationNeeded logs [negotiate] #setRemoteAnswer; failed on every renegotiation from then on, and TracedStreamPeerConnection.setRemoteDescription emits setRemoteDescription.error each time â which is exactly the tracing signal this PR is trying to sharpen.

Suggest making per-candidate failures non-fatal and dropping the offender:

```dart
final result = await pc.setRemoteDescription(sd);
while (_pendingCandidates.isNotEmpty) {
  final candidate = _pendingCandidates.removeAt(0);
  try {
    await pc.addCandidate(candidate);
  } catch (e, stk) {
    // A rejected candidate must not fail the negotiation.
    _logger.w(() => '[setRemoteDescription] dropping candidate: $e');
  }
}
return Result.success(result);
```

That also removes the need for the defensive List.of copy (the lock already guarantees nothing can be appended during the flush) and the O(nÂ²) remove() search. If you'd rather keep failures fatal, at least separate the two try blocks so a candidate error doesn't masquerade as an SDP error.

renefloor (commented):
One more thing that's worth fixing: 

The new tests don't actually reproduce the race they claim to cover (peer_connection_renegotiation_test.dart:461)

The gate is on setRemoteDescription, but the stranding requires addIceCandidate's getRemoteDescription null-read to land after the flush. As written on pre-fix code: setRemoteDescription awaits the gate → addIceCandidate reads null and buffers → gate releases → flush finds the candidate and applies it. So addedCandidates == [candidate] passes without the fix; only results[1].isSuccess fails, which is testing the enum change, not the race. The comment "regardless of which operation acquires the lock first" also isn't exercised — with the lock, ordering is deterministic by call order.

Add a getRemoteDescriptionGate to the fake and hold addIceCandidate there while setRemoteDescription runs to completion. That test fails on main and passes here — which is what makes it a regression test.

renefloor (approved):

PR #1288, opened 2026-07-22
fix(llc): renegotiate unacknowledged transceivers

Brazol:
This PR fixes an issue where republishing could reuse a cached publisher transceiver without renegotiating, even if the previous SetPublisher negotiation failed. Publisher transceivers are now marked as negotiated only after a successful SFU negotiation. If a cached transceiver was never acknowledged by the SFU republishing reuses it but triggers a new negotiation so the track is properly announced.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling for republishing/unmuting to ensure required publisher renegotiation occurs and stale cached transceivers aren’t reused.
  * Added more robust rollback/recovery when negotiation or publisher updates fail.
  * Fixed a `FormatException` caused by non-ASCII client/device info by sanitizing `X-Stream-Client` header values.
  * Updated signaling stall recovery to use a faster reconnection strategy.
* **Improvements**
  * Exposed whether the peer connection is currently reconnecting.
* **Tests**
  * Expanded unit test coverage for transceiver negotiation, republishing, unmuting, and reconnect safety.


packages/stream_video/lib/src/call/session/call_session.dart
```diff
@@ -1139,15 +1142,21 @@ class CallSession extends Disposable {
             _logger.w(
               () => '[negotiate] #setRemoteAnswer; failed: $ansResult',
             );
+            return Result<void>.error(
+              'Failed to set remote answer: ${ansResult.getErrorOrNull()}',
+            );
           }
+
+          rtcManager?.transceiversManager.markNegotiated();
         }
```
renefloor, line 1165: `rtcManager?.transceiversManager.markNegotiated();` sits inside the if (pubResult.data.hasSdp()) block. If SetPublisher succeeds but omits the SDP, the tracks were announced and acknowledged, yet nothing gets marked negotiated. Every subsequent unmute/republish then forces another renegotiation — which again returns no SDP, again marks nothing — so you pay an extra offer/SetPublisher round trip on every unmute for the rest of the session. 

The `markNegotiated` should be after the `if` statement.

If you deliberately mean "only once the answer is applied", then reword the field's doc comment — it currently says "acknowledged by the SFU through a completed publisher negotiation", which the no-SDP success satisfies.

renefloor (approved):

PR #1289, opened 2026-07-23
fix(llc): sanitize header string

### LINKED ISSUE #1283: Non-ASCII app name causes Invalid HTTP header field value in X-Stream-Client
Hello

After updating `stream_video_flutter` to `1.4.1`, Stream Video requests fail when the application display name contains non-ASCII characters.

The SDK reads the application name from `PackageInfo.appName` and includes it directly in the `X-Stream-Client` HTTP header.

For an application named `Giggó` Dart's HTTP client rejects this header before the request is sent:
```
FormatException: Invalid HTTP header field value:
  "stream-video-flutter-v1.4.1|app=[DEV] Giggó|app_version=2.5.1|os=android 16|
  device_model=Nothing A142"
```
 
### Steps to reproduce

1. Set the Android or iOS application display name to a value containing a non-ASCII character, for example Giggó.
2. Initialize StreamVideo.
3. Make a coordinator API request, such as creating or joining a call.
4. The request fails while setting the X-Stream-Client header.

### Expected behavior

The SDK should ensure that X-Stream-Client contains only valid HTTP header characters.

Values collected from the platform, including:

  - application name
  - device model
  - OS name and version
  - browser name

should be sanitized or encoded before being added to the header.

Alternatively, instead of collecting application and device metadata automatically, the SDK could expose a public interface allowing integrators to explicitly provide the metadata they want to send:

  ```dart
  StreamVideoOptions(
    clientMetadata: StreamVideoClientMetadata(
      appName: 'Giggo',
      appVersion: appInfo.version,
      deviceModel: null,
    ),
  );
```

package_info_plus and device_info_plus would not be needed as SDK dependency then.

Brazol:
Fixes https://github.com/GetStream/stream-video-flutter/issues/1283
Resolves FLU-627

Fixed a `FormatException` when sending requests if the application name (or other device/app info) contains non-ASCII characters. Values included in the `X-Stream-Client` header are now sanitized to valid header characters.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed request failures caused by non-ASCII or invalid characters in application and device information.
  * Improved WebSocket connection reliability by safely encoding client information in connection URLs.
  * Sanitized client metadata to produce valid header values.

* **Tests**
  * Added coverage for metadata formatting, non-ASCII characters, and separator handling.

### DIFF
--- a/packages/stream_video/CHANGELOG.md
+++ b/packages/stream_video/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### 🐞 Fixed
+
+- Fixed a `FormatException` when sending requests if the application name (or other device/app info) contains non-ASCII characters. Values included in the `X-Stream-Client` header are now sanitized to valid header characters.
+
 ## 1.4.2
 
 ### ✅ Added
--- a/packages/stream_video/lib/src/coordinator/open_api/coordinator_ws.dart
+++ b/packages/stream_video/lib/src/coordinator/open_api/coordinator_ws.dart
@@ -34,7 +34,7 @@ String _buildUrl(String baseUrl, String apiKey) {
   return '$baseUrl'
       '?api_key=$apiKey'
       '&stream-auth-type=jwt'
-      '&X-Stream-Client=$xStreamClientHeader';
+      '&X-Stream-Client=${Uri.encodeQueryComponent(xStreamClientHeader)}';
 }
 
 class CoordinatorWebSocket extends StreamWebSocket implements HealthListener {
--- a/packages/stream_video/lib/src/sfu/ws/sfu_ws.dart
+++ b/packages/stream_video/lib/src/sfu/ws/sfu_ws.dart
@@ -48,7 +48,7 @@ class SfuWebSocket extends StreamWebSocket implements HealthListener {
           .toString();
     }
     final finalWsEndpoint =
-        '$sfuWsEndpoint?X-Stream-Client=$xStreamClientHeader&attempt=${++sessionSeq}&cid=$cid&user_id=$userId&api_key=$apiKey&user_session_id=$sessionId';
+        '$sfuWsEndpoint?X-Stream-Client=${Uri.encodeQueryComponent(xStreamClientHeader)}&attempt=${++sessionSeq}&cid=$cid&user_id=$userId&api_key=$apiKey&user_session_id=$sessionId';
     streamLog.i(tag, () => '<factory> wsEndpoint: $wsEndpoint');
     streamLog.i(tag, () => '<factory> sfuWsEndpoint: $sfuWsEndpoint');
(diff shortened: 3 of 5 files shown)


renefloor (commented):
🔴 Sanitizing to ASCII isn't enough — the value is also interpolated raw into WebSocket URLs

xStreamClientHeader is used unencoded as a query parameter in two places:

- packages/stream_video/lib/src/sfu/ws/sfu_ws.dart:51
- packages/stream_video/lib/src/coordinator/open_api/coordinator_ws.dart:37

The sanitizer keeps #, &, ?, % — all URL-significant. I verified the impact with Uri.parse on the real SFU URL shape:

appName "Tag #1"    → query={X-Stream-Client: ...|app=Tag }
                      ⚠️ attempt, cid, user_id, api_key, user_session_id ALL DROPPED
appName "Tom & Jerry" → query={X-Stream-Client: ...|app=Tom , " Jerry": "", attempt: 1}
                      ⚠️ client header truncated + junk param injected

A # in the app name silently breaks the SFU handshake entirely. This is the same class of bug the PR is fixing (unusual app names), so it belongs in this PR rather than a follow-up. Preferred fix at the call sites:

'&X-Stream-Client=${Uri.encodeQueryComponent(xStreamClientHeader)}'

That also fixes the pre-existing unencoded space in os=android 16. If you'd rather keep the change local to the sanitizer, restrict the allowlist instead of using a broad printable range — but encoding at the URL boundary is the correct layer.

🟡 CHANGELOG heading should be ## Upcoming

The repo convention is ## Upcoming (see 08495312, 32a1181e, 2b879084); release commits rewrite that heading to the version number. ## Unreleased will be missed by that flow.

🟡 Empty-after-sanitization values leave malformed segments

Nothing handles a value that sanitizes to '':

- App named e.g. 日本語アプリ → app= (empty segment, no signal, still sent)
- osVersion sanitizing to empty → 'os=${name} ' with a trailing space, since the os= branch interpolates unconditionally and .trim() is applied per-value, not to the segment

Suggest treating empty-after-sanitize as absent:

String? _sanitized(String? value) {
  if (value == null) return null;
  final result = /* filter */ .trim();
  return result.isEmpty ? null : result;
}
…then the existing if (x case final v?) patterns handle omission naturally, and the os switch can key on the sanitized version.

🔵 Minor

- Per-call cost / repetition: the getter is recomputed on every HTTP request and WS connect, re-scanning up to 5 strings each time, and _sanitizeHeaderValue appears at 6 call sites. Sanitizing once — in VideoEnvironmentCollector._collect() or the VideoEnvironment constructor — would be cheaper, remove the repetition, and benefit non-header consumers too. Negligible perf impact either way; mostly a readability win.
- codeUnits vs runes: correct as written (surrogate halves are both > 0x7E), just worth a comment so it doesn't look accidental.

renefloor (approved):

PR #1290, opened 2026-07-24
feat(llc): speaking while muted (android, ios, web)

Brazol:
## Summary by CodeRabbit

* **New Features**
  * Speaking-while-muted detection now works on iOS, macOS, and web.
  * Added optional control to keep the microphone track active when muted.
  * Microphone detection adapts when the selected input device changes.
* **Bug Fixes**
  * Resuming suspended audio now preserves the microphone’s mute state.
  * Updating audio bitrate settings no longer unintentionally unmutes participants.
* **Documentation**
  * Updated release documentation with the new microphone controls and platform support.


packages/stream_video/lib/src/webrtc/rtc_manager.dart
```diff
@@ -327,14 +327,39 @@ class RtcManager extends Disposable {
   ) async {
     _defaultAudioConstraints = constraints;
 
-    final localAudioTracks = tracks.values.whereType<RtcLocalAudioTrack>();
+    // Only live tracks are cycled. An already-muted track must stay muted —
```
renefloor, line 330: **Soft mute + `setAudioBitrateProfile` while muted silently does nothing.**

This comment's premise — "it picks the new constraints up from `_defaultAudioConstraints` when it is recreated on unmute" — only holds for hard mute. A `stopTrackOnMute: false` track takes the `track.enable()` branch at the end of `unmuteTrack`, which neither recreates the track (so the new `AudioConstraints` never apply) nor calls `_updateAudioPublishOptions` (so the SFU-side bitrate profile and stereo flag are never updated either).

Failure scenario: iOS user on `stopTrackOnMute: false` mutes, the app calls `setAudioBitrateProfile(musicHighQuality)`, user unmutes -> mono 1-channel capture with AEC/NS/AGC still on, and the SFU still on the old profile. Nothing recovers it until a hard mute or a device switch.

Before this PR the enable-only branch was effectively dead for the mic (nothing could set `stopTrackOnMute: false` on an audio track), so this PR is what makes it reachable. Suggest either recreating soft-muted audio tracks on unmute when `_defaultAudioConstraints` changed since capture started, or applying `_updateAudioPublishOptions` in the enable-only branch too.

pubspec.lock
```diff
@@ -2246,10 +2246,10 @@ packages:
     dependency: transitive
     description:
       name: vector_math
-      sha256: f36f9f3be64c6198714492bb455c11056e33e2f85d9a0b676a48301e44fdcf47
+      sha256: d530bd74fea330e6e364cda7a85019c434070188383e1cd8d9777ee586914c5b
       url: "https://pub.dev"
     source: hosted
-    version: "2.4.2"
+    version: "2.2.0"
```
renefloor, line 2252: **Unrelated transitive downgrades.** This lock diff drops five packages: `intl` 0.20.3->0.20.2, `matcher` 0.12.20->0.12.19, `meta` 1.19.0->1.18.0, `test_api` 0.7.12->0.7.11, and `vector_math` 2.4.2->**2.2.0** (two minor versions).

Looks like `pub get` ran on an older Flutter SDK than the one that produced the lockfile on `main`. Worth re-resolving on the CI Flutter version so the lock diff is limited to `stream_webrtc_flutter`.

packages/stream_video/lib/src/call/session/call_session.dart
```diff
@@ -1253,6 +1253,15 @@ class CallSession extends Disposable {
     onSuspendedAudioTrackRecorded(track.trackId);
   }
 
+  /// Whether [track] is the local microphone track and the SDK currently
+  /// considers it muted.
+  bool _isMutedLocalMicrophoneTrack(RtcTrack track) {
```
renefloor, line 1258: **`?? false` makes "no local participant" mean "muted".**

`!(localParticipant?.isAudioEnabled ?? false)` returns `true` when `localParticipant` is null, so `resumeSuspendedAudioTracks` leaves the mic disabled. If the local participant hasn't landed in state yet when a resume runs, the user comes out of the suspension with a dead mic and no SDK state saying so.

`?? true` fails the other way (briefly on-air), which I assume is why you picked this — fair. But could the null case at least log at `w` level? Right now it's indistinguishable from a legitimate mute, which makes it undiagnosable from a customer log. Was the null branch deliberate?

packages/stream_video/lib/src/audio_processing/speaking_while_muted_recognition.dart
```diff
@@ -97,32 +151,90 @@ class SpeakingWhileMutedRecognition
...
   }
 
-  Future<void> _stop() async {
-    if (!_isActive) return;
-    _isActive = false;
+  Future<void> _stopRecognition() async {
+    if (!_isRunning) return;
+    _isRunning = false;
+    _logger.i(() => '[stopRecognition] no args');
     state = const SpeakingWhileMutedState._(isSpeakingWhileMuted: false);
```
renefloor: **This can write state to an unmounted notifier.** `dispose()` doesn't await `_transitions`, and this write is unguarded.

Reachable path on web: mute -> `_startRecognition` blocks in `getUserMedia` (permission prompt) -> unmute enqueues `_stop` -> user leaves the call. `dispose()` sets `_disposed`, awaits two fast operations, calls `super.dispose()`. `getUserMedia` then resolves, the queued `_stop` transition runs, `_isRunning` is still `true`, and this line hits `state_notifier`'s `assert(_debugIsMounted())` -> `StateError` in debug builds.

One-line fix: `if (_disposed) return;` above this, or `await _transitions` in `dispose()`.

dogfooding/lib/screens/call_screen.dart
```diff
@@ -315,6 +315,9 @@ class _CallScreenState extends State<CallScreen> {
                           call: call,
                           disabledMicrophoneBackgroundColor:
                               AppColorPalette.appRed,
+                          // Keep the track alive on mute so speaking-while-
```
renefloor, line 318: **Unconditional `false` costs a duplicate capture on web.** The flag is only *needed* on iOS/macOS, but `AudioRecognitionWebAudio` always opens its own `getUserMedia` — so on web this leaves the published track alive *and* a second capture of the same mic open. Chrome tolerates that; Safari/Firefox are less reliable on a duplicate same-device capture, and it doubles AEC work.

Since integrators will copy this example verbatim, consider gating it:

```dart
stopTrackOnMute: CurrentPlatform.isIos || CurrentPlatform.isMacOS ? false : null,
```

Separately — could the web recognizer reuse the existing local audio track's `MediaStream` when one is alive, the way the JS SDK's `createSoundDetector` does, rather than opening its own?

packages/stream_video/lib/src/audio_processing/audio_recognition_web_audio.dart
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  void _stopStream(web.MediaStream stream) {
+    for (final track in stream.getTracks().toDart) {
+      track.stop();
+    }
+  }
+}
+
+class WebAudioRecognitionConfig {
```
renefloor, line 193: **This config is unreachable from outside the package.** Neither `AudioRecognitionWebAudio` nor `WebAudioRecognitionConfig` is exported from `stream_video.dart`, and `createPlatformAudioRecognition` never passes a config — so `audioLevelThreshold`, `detectionInterval`, and `fftSize` can't be tuned by integrators.

`deviceId` (line 200) is fully dead: only `deviceIdProvider` is ever wired up. Either export the config — the class docs already point people at "supply a custom `AudioRecognition`", so the extension point is public in spirit — or drop `deviceId`.

packages/stream_video/lib/src/audio_processing/audio_recognition_web_audio.dart
```diff
@@ -0,0 +1,225 @@
...
+
+    _stream = stream;
+    final audioContext = web.AudioContext();
+    _audioContext = audioContext;
+
+    final source = audioContext.createMediaStreamSource(stream);
+    final analyser = audioContext.createAnalyser()..fftSize = config.fftSize;
+    source.connect(analyser);
+
+    // Detection can starts outside a user gesture (a moderator mute, or
```
renefloor: Nit: typo — "Detection can starts" -> "Detection can start".

packages/stream_video/lib/src/webrtc/rtc_manager.dart
```diff
@@ -1080,9 +1110,14 @@ extension PublisherRtcManager on RtcManager {
     _logger.i(() => '[publishAudioTrack] track: $audioTrack');
     tracks[audioTrack.trackId] = audioTrack;
     var updatedTrack = audioTrack.copyWith(stopTrackOnMute: stopTrackOnMute);
-
     var needsRenegotiation = false;
```
renefloor, line 1113: Minor: this unmute and the `reconcileAppleAdmMicrophoneMute()` at the end of `publishAudioTrack` are redundant for correctness — the reconcile covers this case. If the point is to unmute capture *before* negotiation rather than after, a one-line comment saying so would stop someone deleting the apparent duplicate later.

packages/stream_video/test/src/audio_processing/audio_recognition_webrtc_test.dart
```diff
@@ -0,0 +1,136 @@
...
+          config: const AudioRecognitionConfig(
+            speechTimeout: Duration(milliseconds: 50),
+            maxSpeechDuration: Duration(milliseconds: 200),
+          ),
+          settle: const Duration(milliseconds: 400),
+        );
+
+        expect(emitted, [true, false, true, false]);
+      },
+    );
```
renefloor: These tests drive real `Timer`s through `Future.delayed`, with margins as tight as a 50ms step against a 200ms `maxSpeechDuration`. On a loaded CI runner that's a flake source — `fakeAsync` would make the whole group deterministic.

renefloor (commented):
Solid work overall. The ADM-mute reconcile design is the right call — treating the ADM as external state that drifts and re-asserting it at every capture-restarting boundary (publish, unmute, device switch, suspend-resume) is much more robust than trying to keep it in sync imperatively. Test coverage is unusually thorough, including the awkward paths that need the platform plugin, and the two suspend/mute race fixes found along the way are genuine bugs.

Reviewed at `e1f7b776`. Inline comments below, roughly worst-first. Two things worth calling out up front:

- **#1 (`changeDefaultAudioConstraints`)** is the one I'd want resolved before merge — soft mute makes `setAudioBitrateProfile`-while-muted a silent no-op.
- **#2 (`pubspec.lock`)** looks like a stray resolution on an older Flutter SDK, unrelated to this feature.

Also: CI (`analyze`, `build`, the package test jobs) was still pending when I looked, so I haven't verified the suite is green.

Nice cleanup removing the two commented-out `token_*.dart` files in `e1f7b776`.

renefloor (approved):

PR #1291, opened 2026-07-27
fix(llc): publisher track mid mismatch

Brazol:
Fixes publisher-negotiation bugs that could leave a track "published" but sending no/wrong media.
The announced track mid was read from a cached transceiver and could drift from the offer after a renegotiation/retry, so the SFU couldn't match the track to its codec. It's now resolved from the live peer-connection state to always match the offer being sent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed track announcement during renegotiation and republish retries to avoid unresolved/stale MIDs, preventing missing or dropped tracks.
  * Improved MID resolution and transceiver matching, including safer fallback behavior when MID lookup isn’t possible.
  * Enhanced signaling robustness by properly handling remote-description/ICE-candidate serialization and sanitizing client header values with non-ASCII characters.
* **Tests**
  * Added/expanded unit and session tests for MID resolution precedence, reconnect/rollback behavior, and negotiated-MID tracking/edge cases.


renefloor (commented):
🔴 The recovery Result is thrown away — _onRenegotiationNeeded always returns success

Every return the PR adds is inside _negotiationLock.synchronized(() async { ... }), so it returns from the closure, not from _onRenegotiationNeeded. call_session.dart:1150 unconditionally returns const Result.success(null). This predates the PR (the old Result.error('Failed to create offer') was equally discarded), but the PR is built entirely around returning recovery results, so it needs fixing here.

Consequence — startPublisherConnectionCheck at call_session.dart:638-646:

final result = await _onRenegotiationNeeded(publisher);
if (result.isFailure && !_isLeavingOrClosed) { ... rejoin ... }   // dead branch

The have-local-offer stall path can never escalate to rejoin. That's one of the two stalls this PR claims to fix, and it's still broken. Same for the onIceRestart failure logging at line 974.

return await _negotiationLock.synchronized<Result<void>>(() async {
  ...
  return const Result.success(null);   // explicit success at the end of the closure
});

Note the interaction once fixed: _recoverFailedPublisherNegotiation returning Result.success(null) after firing onReconnectionNeeded becomes load-bearing — it's what stops line 645 from firing a second rejoin. That's a defensible choice but it's currently implicit; please comment it.

🟠 getTransceivers() can throw, and neither call site is protected

getAnnouncedTracks is invoked at call_session.dart:1111 — before the try at line 1122 — and createOffer has already set the local description by then. The native impl throws a bare String on PlatformException, and RTCRtpSenderNative.fromMap does Map<dynamic, dynamic> trackMap = map['track']; (non-nullable) — a TypeError if the platform omits track. Any throw escapes _onRenegotiationNeeded entirely: no rollback, no rejoin, publisher wedged in have-local-offer. That is precisely the failure mode this PR set out to eliminate.

The reconnect path is worse: getReconnectDetails (call_session.dart:172) awaits getAnnouncedTracksForReconnect, which now also calls getTransceivers(). On a publisher pc that's already closing during a rejoin, a throw here aborts the whole reconnect rather than degrading.

Wrap it:

final liveTransceivers = await publisher?.pc.getTransceivers().onError((_, __) => []) ?? [];

…or move getAnnouncedTracks inside the existing try in _onRenegotiationNeeded.

🟡 Dropping the type == kind check makes the SDP fallback fail closed and silent

Old predicate: m['type'] == track.kind && ((m['msid'] as String?)?.contains(track.id!) ?? true) — the ?? true meant "no msid on this m-section → match on kind alone." New predicate requires msid present and containing the trackId, else null → the track is dropped from the announce and therefore never published.

I think failing closed is the right call (announcing a wrong mid is worse than announcing nothing), but two asks:

- Keep m['type'] == kind as an additional guard on the msid match. It's free and protects against the msid appearing in a recycled/inactive m-section that .reversed might land on.
- The drop is only a _logger.w. The user-visible symptom is "my camera isn't being published" with no error surfaced anywhere. Add a _tracer.trace so it shows up in telemetry — you already trace publisherConnectionCheckStalled for comparable conditions.

🟡 No tests

resolveTrackMid is the most testable function in this PR — pure over a transceiver list plus an SDP string, and test/src/webrtc/peer_connection_renegotiation_test.dart already has the mock-PC scaffolding. Worth covering:

- live mid wins over SDP
- empty live mid → SDP msid fallback resolves
- .reversed picks the last matching m-section (the recycled-m-line case that motivated this fix)
- msid absent → null (pins the new fail-closed contract)
- mediaTrack.id == null → null

Plus one on the new escalation: empty tracksInfo + no remote description → onReconnectionNeeded(rejoin).

🔵 Minor

- Dead code: TransceiverManager.indexOf (transceiver_cache.dart:114) and its backing _transceiverOrder list are now unused — rtc_manager.dart:703 was the only caller. Both should go in this PR.
- resolveTrackMid(TransceiverCache cache, ...) reads only cache.track.mediaTrack.id. Taking String? trackId would be a more honest signature and trivially testable.
- _recoverFailedPublisherNegotiation logs 'initial publisher offer failed' for all three call sites, including setPublisher failure. Pass a reason string in.
- parse(sdp) runs per unresolved track inside the loop. Rare in practice (step 1 usually wins, since createOffer sets the local description before mids are read), so low priority — but hoisting the parse into the two getAnnouncedTracks* callers would remove it entirely.

renefloor (approved):

PR #1303, opened 2026-08-13
fix(llc): fixes around audio output on web

Brazol:
## Summary by CodeRabbit

* **New Features**
  * Added web audio playback-blocked status reporting and recovery for browser autoplay restrictions and unexpected pauses.
  * Added an API to resume blocked web audio playback.
  * Added tracing for audio playback recovery and audio output-device changes.

* **Bug Fixes**
  * Improved remote audio playback recovery and audio output-device switching.
  * Preserved audio output settings when remote audio is unmuted.
  * Preserved web audio resources when remote tracks are unpublished.
  * Audio sink changes now complete asynchronously and report routing failures.
  * Improved Opus DTX and RED negotiation.


packages/stream_video/lib/src/call/stats/trace_tag.dart
```diff
@@ -111,4 +112,12 @@ abstract class TraceTag {
...
       'navigator.mediaDevices.getDisplayMedia';
   static const String getUserMedia = 'navigator.mediaDevices.getUserMedia';
+
+  // Remote audio playback (web only).
+  static const String resumeAudio = 'resumeAudio';
+  static const String mediaPlaybackPaused = 'mediaPlayback.paused';
+  static const String mediaPlaybackBlocked = 'mediaPlayback.blocked';
+  static const String mediaPlaybackRecovered = 'mediaPlayback.recovered';
+  static const String mediaPlaybackRecoverGiveUp =
+      'mediaPlayback.recover.giveUp';
```
renefloor: I think for v2 we should for webrtc_dart and put all web only stuff in there when possible

packages/stream_video/lib/src/webrtc/rtc_audio_api/rtc_audio_html.dart
```diff
@@ -77,3 +157,177 @@ web.HTMLDivElement findOrCreateAudioContainer() {
...
+        }
+        return;
+      }
+
+      if (_attempt >= _maxRecoveryAttempts) {
+        _trace(TraceTag.mediaPlaybackRecoverGiveUp, {
+          'trackId': trackId,
+          'kind': 'audio',
+          'attempts': _attempt,
+          'error': '$e',
```
renefloor: **Blocking: give-up permanently kills the track, with no way back.**

After this early return, `_degraded` is never reset (only `_handlePlaying` clears it), so `_handlePauseOrSuspend`'s `if (_disposed || _degraded) return;` swallows every subsequent `pause`/`suspend` — no trace, no retry. And the track isn't in `_blockedTrackIds`, so `resumeAudioPlayback()` won't reach it either. That participant is silent for the rest of the call.

The backoff sums to ~27s across the 10 attempts, so a transient ~30s window (device switch, tab throttling) is enough to burn the whole budget and hit this.

Suggest resetting `_degraded = false; _attempt = 0;` before returning here so a later `pause` can start a fresh cycle — or keep a separate `_gaveUp` flag that `resumeAudioPlayback()` also clears.

packages/stream_video/lib/src/webrtc/rtc_audio_api/rtc_audio_html.dart
```diff
@@ -40,19 +81,58 @@ void stopAudio(String id) {
...
+  _trace(TraceTag.resumeAudio, {'blockedTracks': _blockedTrackIds.length});
 
-  if (audioElement is web.HTMLAudioElement &&
-      audioElement.hasProperty('setSinkId'.toJS).toDart) {
-    audioElement.setSinkId(deviceId);
+  final blocked = List<String>.from(_blockedTrackIds);
+  for (final trackId in blocked) {
+    _blockedTrackIds.remove(trackId);
+    await _watchdogs[trackId]?.play();
   }
```
renefloor: Two things here:

**1. `_attempt` is never reset on a manual resume.** Each blocked `play()` still does `_attempt += 1` before the `NotAllowedError` early-return, so the counter climbs with every user retry. If the 11th gesture-driven attempt fails with anything *other* than `NotAllowedError`, it hits the give-up ceiling instantly instead of retrying. Worth resetting `_attempt = 0` when resuming from a gesture.

**2. Sequential awaits (agreeing with CodeRabbit).** The loop waits for each `play()` before starting the next. Start them all, then `Future.wait`, so every call lands inside the gesture.

packages/stream_video/lib/src/webrtc/rtc_audio_api/rtc_audio_html.dart
```diff
@@ -77,3 +157,177 @@ web.HTMLDivElement findOrCreateAudioContainer() {
...
+      streamLog.w(
+        _tag,
+        () =>
+            'Playing audio for track $trackId failed '
+            '(attempt $_attempt): $e',
+      );
+
+      // The autoplay policy will keep rejecting until a user gesture, so stop
+      // burning attempts and wait for [resumeAudioPlayback].
+      if ('$e'.contains('NotAllowedError')) {
```
renefloor: This string match is the single gate on the entire feature — if it misses, `isAudioPlaybackBlocked` never goes true and the new API is dead code, while the watchdog burns 10 pointless retries.

It happens to work on dart2js because `DOMException.toString()` is `"${name}: ${message}"`, but that depends on the interop error surviving as a JS object with a live `toString`, which I wouldn't assume holds under dart2wasm. Reading the property is cheap and exact:

```dart
final name = e is jsutil.JSObject
    ? e.getProperty<jsutil.JSString?>('name'.toJS)?.toDart
    : null;
if (name == 'NotAllowedError') { ... }
```

packages/stream_video/lib/src/webrtc/rtc_media_device/rtc_media_device_notifier.dart
```diff
@@ -296,6 +300,19 @@ class RtcMediaDeviceNotifier {
     return rtc.Helper.resumeAudioPlayout();
   }
 
+  /// Whether the browser's autoplay policy is currently blocking playback of at
+  /// least one remote audio track.
+  ///
+  /// Web only; always `false` on other platforms.
+  bool get isAudioPlaybackBlocked => rtc_audio.isAudioPlaybackBlocked;
```
renefloor: **`isAudioPlaybackBlocked` has no change notification.**

The whole point of this API is "show an unblock-audio button", but it's a plain getter over a global `Set` — an app has no way to learn when it flips, short of polling. Everything else on this notifier is a stream (`onDeviceChange`, `speechActivityStream`), so this is inconsistent too.

This should be a `Stream<bool>` / `BehaviorSubject`. Much cheaper to change now than after it ships as public API.

packages/stream_video/lib/src/webrtc/rtc_audio_api/rtc_audio_html.dart
```diff
@@ -77,3 +157,177 @@ web.HTMLDivElement findOrCreateAudioContainer() {
...
+      unawaited(play());
+    });
+  }
+
+  String? _computeSkipReason() {
+    if (_disposed) return 'disposed';
+    if (element.srcObject == null) return 'noSrc';
+    if (element.ended) return 'ended';
+    if (_blockedTrackIds.contains(trackId)) return 'blocked';
+    if (element.readyState < _haveCurrentData) return 'notReady';
```
renefloor: `notReady` short-circuits recovery before anything is traced. Worth confirming on a real device that the target scenario (Bluetooth profile switch pausing the element) leaves `readyState >= 2`.

If it drops below that, the primary bug this PR fixes goes unrecovered *and* untraced — the same silent failure, in a new place. Not something I can settle from the code; needs a manual check.

packages/stream_video/lib/src/webrtc/rtc_audio_api/rtc_audio_html.dart
```diff
@@ -40,19 +81,58 @@ void stopAudio(String id) {
...
+    await audioElement.setSinkId(deviceId).toDart;
+    streamLog.i(_tag, () => 'Set sink id for track $id to $deviceId');
+    _trace(TraceTag.setSinkId, {'trackId': id, 'deviceId': deviceId});
+  } catch (e) {
+    streamLog.e(_tag, () => 'Failed to set sink id for track $id: $e');
+    _trace(TraceTag.setSinkId, {
+      'trackId': id,
+      'deviceId': deviceId,
+      'error': '$e',
+    });
```
renefloor, line 174: Agreeing with CodeRabbit here: the rejection is swallowed, so `RtcRemoteTrack.setSinkId` still stores the rejected `audioSinkId` and `RtcManager.setAudioOutputDevice` returns `Result.success`. State then claims a device the browser refused. Propagate the failure, or return something the caller can branch on.

renefloor (commented):
The diagnosis is right and the shape of the fix is sound — `autoplay` alone genuinely does fail silently under autoplay policy, and driving `play()` explicitly with a watchdog is the correct remedy.

The two smaller fixes bundled in are both real:

- `.toList()` in `setAudioOutputDevice` is now **required**, not cosmetic — the loop awaits, so `onRemoteTrack` can insert into `tracks` mid-iteration and blow up the lazy iterable.
- `disposeWebAudioPlayer: false` on unpublish is correctly bounded: participant-left and call teardown both go through `unpublishTrack` → `stop()` with the default `true`, so elements still get cleaned up.

One thing I'd want fixed before merge (inline): after the watchdog gives up, the track is permanently silent with no path back — not even a user gesture. The rest are worth-addressing / nits.

Also: patch coverage is 10%. The DOM parts are awkward to test, but `_backoff()` is pure and the `_degraded`/`_computeSkipReason` transitions would be testable if the watchdog took a small element abstraction. Given that the one blocking finding *is* a state-machine bug, some coverage there would pay for itself.

CHANGELOG correctly under `## Upcoming` ✅

packages/stream_video/lib/src/webrtc/rtc_audio_api/rtc_audio_html.dart
```diff
@@ -40,31 +89,114 @@ void stopAudio(String id) {
...
+
+bool checkIfAudioOutputChangeSupported() {
+  final cached = _audioOutputChangeSupported;
+  if (cached != null) return cached;
+
+  // Detect the capability by looking the method up rather than calling it:
+  // `setSinkId` returns a promise, so invoking it would route audio and leave
+  // a rejection unhandled instead of reporting support.
+  final element = web.document.createElement('audio');
+  return _audioOutputChangeSupported =
```
renefloor: **Blocking: this is the `format:verify` failure.**

`analyze` is red on the branch, and this is the only cause — `dart format` reports 1 changed file out of 869, and it's this statement:

```dart
  return _audioOutputChangeSupported = element
      .hasProperty('setSinkId'.toJS)
      .toDart;
```

```bash
dart format packages/stream_video/lib/src/webrtc/rtc_audio_api/rtc_audio_html.dart
```

Unrelated to the change itself — the rewrite of this function is a real improvement. The old version *invoked* `setSinkId('default')` to probe for support, which actually routed audio and left a floating promise rejection; looking the method up instead is both cheaper and side-effect free.

packages/stream_video/CHANGELOG.md
```diff
@@ -1,7 +1,15 @@
 ## Upcoming
 
+### 🔄 Changed
+
+- [Web] `RtcRemoteTrack.setSinkId` is now asynchronous (`Future<RtcRemoteTrack>` instead of `RtcRemoteTrack`) and throws when the browser cannot route the track to the requested device.
```
renefloor, line 10: **Blocking: missing `### ✅ Added`.**

Three new members land on `RtcMediaDeviceNotifier`, which is public API — it's exported from `stream_video.dart:57`:

- `isAudioPlaybackBlocked`
- `isAudioPlaybackStalled`
- `resumeAudioPlayback()`

The `## Upcoming` block documents the `setSinkId` signature change and all four fixes but never mentions the new API. 1.4.2 sets the precedent with its own `### ✅ Added` section.

While you're in here: the `copyWith` change forwarding `transceiver` isn't listed either, and it's a real fix independent of the audio work. Previously *any* `copyWith` on an `RtcRemoteTrack` dropped the transceiver, which silently no-op'd the `removeTrack` in `unpublishTrack` — worth its own `### 🐞 Fixed` line rather than riding along unmentioned.

packages/stream_video/lib/src/webrtc/rtc_media_device/rtc_media_device_notifier.dart
```diff
@@ -296,6 +300,26 @@ class RtcMediaDeviceNotifier {
...
```
renefloor: **Still open, and narrower than I said last round: there is no state update at all.**

I traced every write site. All mutations of `_blockedTrackIds` / `_stalledTrackIds` live inside `rtc_audio_html.dart`, and nothing in that file touches `stateManager`, `CallStateNotifier`, or any emitter. These two getters are the only readers, and they're synchronous reads of a page-global `Set`.

So the signal has exactly two outlets:

1. `streamLog` — a log line.
2. `_trace(...)` → this notifier's `Tracer` → drained by `SfuStatsReporter` (`sfu_stats_reporter.dart:247`) → uploaded to the SFU in `SendStatsRequest`.

That second one is the part I'd push back on. The only real consumer of the blocked/stalled signal is **our own backend telemetry** — we'll see on a dashboard that a user's audio was blocked, while the app whose user is sitting in silence has to poll a getter to find out. The five new `mediaPlayback.*` tags are good observability aimed at the wrong audience.

The precedent for the right shape is already in this same feature area: the selected output device *is* call state — `CallState.audioOutputDevice`, written through `stateManager.participantSetAudioOutputDevice()` (`state_participant_mixin.dart:148`), observable via `call.state`. Grepping `call_state.dart` for `audio` returns nothing, so there's no playback-health field to hang this on yet.

Concretely: route the blocked/stalled transitions into `CallState` alongside `audioOutputDevice`, so `resumeAudioPlayback()` has a reactive trigger for a "tap to enable sound" affordance. The web-only globals can stay as the mechanism — they just need to notify outward as well as into the tracer.

This is also why I'd keep it pre-merge rather than follow-up: these getters are already public API, so adding the state field later means shipping two ways to ask the same question.

packages/stream_video/lib/src/webrtc/rtc_audio_api/rtc_audio_html.dart
```diff
@@ -77,3 +209,209 @@ web.HTMLDivElement findOrCreateAudioContainer() {
...
+  void _scheduleRecovery() {
+    if (_disposed || _pendingTimer != null) return;
+
+    final skipReason = _computeSkipReason();
+    if (skipReason != null) {
+      streamLog.v(
+        _tag,
+        () => 'Recovery skipped for track $trackId ($skipReason)',
+      );
+      return;
```
renefloor, line 441: **`_degraded` is a latch whose exits are incomplete — same shape as the give-up bug, one path over.**

This skip branch returns without touching `_degraded`, and `play()`'s success path (L292-294) doesn't clear it either — only `_handlePlaying` does. So if `_scheduleRecovery` is entered from `play()`'s catch and `_computeSkipReason()` returns `notPaused`, you land in: `_degraded == true`, no timer armed, and the track in neither `_blockedTrackIds` nor `_stalledTrackIds` — so `resumeAudioPlayback()` can't reach it. Every subsequent `pause`/`suspend` is then swallowed by `if (_disposed || _degraded) return` in `_handlePauseOrSuspend`.

Being fair about severity: this is narrow. It needs a non-`NotAllowedError` rejection racing a self-recovery, and a self-recovery normally fires `playing`, which clears the latch. I couldn't construct a scenario I'm confident happens in the field, so I'd call it robustness rather than a live bug — not blocking.

But the reason I'd still fix it is that the reachability argument is the *only* thing keeping it safe, and it's the second time this latch has produced a dead-end. Two lines make it unconditional instead of case-dependent:

- reset `_degraded = false` in this skip branch
- clear `_degraded` (and `_attempt`) on `play()` success at L292-294, rather than relying on the `playing` event to arrive

Then every exit from the recovery machine re-arms it, and no future edit can reintroduce this by adding a new early return.

packages/stream_video/lib/src/call/stats/trace_tag.dart
```diff
@@ -100,6 +100,7 @@ abstract class TraceTag {
       'audioProcessingStateChanged';
   static const String enumerateDevices =
       'navigator.mediaDevices.enumeratedevices';
+  static const String setSinkId = 'navigator.mediaDevices.setSinkId';
```
renefloor: **Wrong namespace.** `setSinkId` is `HTMLMediaElement.setSinkId` — it isn't on `navigator.mediaDevices` at all. The surrounding tags in this block (`enumeratedevices`, `pauseAudioPlayout`, `getUserMedia`) are all genuinely `navigator.mediaDevices` members, so this one is borrowing a prefix that doesn't apply.

Worth getting right in the same pass rather than later: these strings are what dashboards join on across SDKs, so a wrong name is expensive to correct once data has accumulated under it.

packages/stream_video/lib/src/call/stats/trace_tag.dart
```diff
@@ -111,4 +112,12 @@ abstract class TraceTag {
   static const String getDisplayMedia =
       'navigator.mediaDevices.getDisplayMedia';
   static const String getUserMedia = 'navigator.mediaDevices.getUserMedia';
+
+  // Remote audio playback (web only).
+  static const String resumeAudio = 'resumeAudio';
```
renefloor, line 122: **Inconsistent with every neighbouring tag.** These five drop the namespace prefix that the rest of the file carries — `resumeAudio` in particular is bare, and reads oddly next to the existing `navigator.mediaDevices.resumeAudioPlayout` two blocks up, which is a *different* operation.

Not a correctness issue, but same reasoning as the comment above: worth settling the naming before data accumulates. If these are meant to line up with the JS SDK's `MediaPlaybackWatchdog` / `BlockedAudioTracker` tags (the class doc says it mirrors them), matching those names exactly would be the most useful thing here — cross-SDK comparison is the whole point of the tags.

packages/stream_video/lib/src/webrtc/rtc_manager.dart
```diff
@@ -1885,11 +1885,38 @@ extension RtcManagerTrackHelper on RtcManager {
...
 
-      for (final audioTrack in audioTracks) {
-        final updatedTrack = audioTrack.setSinkId(device.id);
+      // Try every track instead of aborting on the first rejection: stopping
+      // half way leaves the switched tracks on the new device and the rest on
+      // the old one, with no way to record both.
+      final applied = <RtcRemoteTrack>[];
+      final failures = <(String, Object, StackTrace)>[];
+
+      await Future.wait(
```
renefloor, line 1894: **This now re-applies to every track on every new remote audio track, and each one awaits a real browser promise.**

`_onRemoteTrackReceived` calls `_applyCurrentAudioOutputDevice()` → `setAudioOutputDevice` → `setSinkId` on all N tracks, so joining an N-party call issues O(N²) `setSinkId` calls. That was already true before this PR, but they used to be fire-and-forget; now each is awaited, so the cost lands as wall time in the join path.

Cheap fix: skip tracks whose `audioSinkId` already equals `device.id`. They still need to go into `applied` so the `applied.isEmpty && failures.isNotEmpty` check at L1914 keeps its meaning — otherwise a call where every track is already on the device would look like a total failure.

The `.toList()` above is correct and load-bearing now, by the way — the body awaits, so `_onRemoteTrack` can insert into `tracks` mid-iteration and invalidate a lazy iterable. Worth a short comment saying so, since it reads as cosmetic.

packages/stream_video/lib/src/webrtc/rtc_audio_api/rtc_audio_html.dart
```diff
@@ -40,31 +89,114 @@ void stopAudio(String id) {
...
+    throw UnsupportedError(
+      'Audio output device change is not supported on this browser.',
     );
-    return;
   }
 
   final audioElement = web.document.getElementById(audioPrefix + id);
+  if (audioElement is! web.HTMLAudioElement) {
+    // Not a failure: the element is created when the track starts, and
+    // `start()` applies the sink id the caller is about to record.
```
renefloor: This comment overstates the guarantee, and it's load-bearing — it's the justification for returning success from a call that routed nothing.

"`start()` applies the sink id the caller is about to record" only holds if the same `RtcRemoteTrack` instance is the one later started. On a republish it isn't: `rtc_manager.dart:296` constructs a fresh `RtcRemoteTrack` with `audioSinkId: null` and overwrites the map entry, dropping the id that was just recorded.

It does work in practice, but via a different mechanism — both `_onRemoteTrackReceived` and `resumeSuspendedAudioTracks` call `_applyCurrentAudioOutputDevice()` immediately after `start()`, which re-derives the device from call state. Worth saying that instead, so the next person doesn't rely on a guarantee `copyWith` doesn't actually provide.

renefloor (commented):
Re-reviewed at `3bbe656`. Everything I flagged last round is genuinely fixed — the give-up path now resets `_degraded`/`_attempt` and routes through the new `_stalledTrackIds` so `resumeAudioPlayback()` can reach it, `resume()` zeroes `_attempt`, the retries go out under one `Future.wait`, `NotAllowedError` is detected by reading the `name` property, and `setSinkId` rejections propagate all the way to `Result.failure`. Dropping the `readyState` gate from `_computeSkipReason` also retires my "needs a manual check on a Bluetooth profile switch" question — a paused element now always enters recovery regardless of buffer state.

Two things block merge, one design question is still open, and the rest are inline.

**Blocking:** `format:verify` is red (one spot, inline), and the CHANGELOG is missing an `### ✅ Added` section for the new public API.

**Still open from last round:** `isAudioPlaybackBlocked` has no change notification. I dug into this further and it's narrower than I first described — see the inline comment. Short version: there is no state update at all, and the only real consumer of the signal is our own backend telemetry.

**Tests.** Still nothing. Two of the seven findings across both rounds have been bugs in the same watchdog state machine, which is the strongest argument for coverage there. `_backoff()` is pure and needs no harness. `RtcManager.setAudioOutputDevice`'s partial-failure branch (all-fail → failure, some-fail → success) needs no DOM at all. And behind a two-member element abstraction (`play()`, `paused`) the `_degraded`/`_computeSkipReason` transitions become ordinary unit tests — including the latch case below, which is exactly the kind of thing that is invisible in review and obvious in a test.

packages/stream_video/lib/src/webrtc/rtc_audio_api/rtc_audio_html.dart
```diff
@@ -77,3 +235,221 @@ web.HTMLDivElement findOrCreateAudioContainer() {
...
+          'readyState': element.readyState,
+          'error': '$e',
+        });
+
+        // Stop retrying this outage, but re-arm the watchdog. Not reported
+        // outwards: the autoplay policy isn't what stopped playback here.
+        _stalledTrackIds.add(trackId);
+        _degraded = false;
+        _attempt = 0;
+        return;
```
renefloor: **Blocking: give-up is terminal again, from the app's side.**

Last round's fix for "give-up permanently kills the track" was `_stalledTrackIds` plus `resumeAudioPlayback()` reaching them. That worked because `isAudioPlaybackStalled` was a public getter. This round removed it and replaced it with `isWebAudioPlaybackBlocked`, which deliberately excludes stalled tracks (the comment right above).

Trace the path after the 10th failed attempt:

1. The element stays paused, so no further `pause`/`suspend` event can fire — "re-arm the watchdog" has no trigger.
2. `_computeSkipReason` (L440) doesn't consult `_stalledTrackIds`, so nothing re-enters recovery either.
3. No signal reaches the app, so nothing prompts a `resumeWebAudioPlayback()` call.

That participant is silent for the rest of the call — the exact failure this PR set out to fix, one path further down.

Two ways out: surface stalled separately in `CallState`, or don't give up at all and keep retrying at the 5s cap. Widening `isWebAudioPlaybackBlocked` to cover it would be wrong — the dartdoc promises it means the autoplay policy specifically.

renefloor (commented):
Reviewed at `472744b5`. Everything from the last two rounds is genuinely fixed: the state flows through `CallState.isWebAudioPlaybackBlocked` instead of a polled getter, `_markPlaying` unifies both recovery observers, the `_degraded` latch has an exit on the skip path, `_attempt` resets on gesture-driven resume, `_isNotAllowedError` reads the property, the same-sink shortcut kills the O(N^2) awaits, and the changelog has its `Added` section. Format is green.

Five things I'd still want settled, one of them a regression introduced by this round's API redesign — the stalled-track path is no longer reachable from the app, so give-up is terminal in practice. Details inline.

packages/stream_video/lib/src/webrtc/rtc_audio_api/rtc_audio_html.dart
```diff
@@ -40,31 +111,118 @@ void stopAudio(String id) {
...
-  } catch (_) {
-    return false;
+  final pending = {
+    ..._blockedTrackIds,
+    ..._stalledTrackIds,
+  }.where(_watchdogs.containsKey).toSet();
+
+  _blockedTrackIds.removeAll(pending);
+  _stalledTrackIds.removeAll(pending);
+  _notifyBlocked();
```
renefloor: **The flag flickers off before you know whether the retry worked.**

This clears `_blockedTrackIds` and notifies *before* the retries run. If the gesture-driven retry still fails, `_blockedTrackIds.add` re-notifies true a moment later. Net UI effect: the "tap to enable sound" button disappears and comes back.

Nothing depends on the early removal — `resume()` calls `play()` directly and never consults `_computeSkipReason`, so the `blocked` skip branch isn't in play here. Either move `_notifyBlocked()` below `await Future.wait(retries)`, or drop the up-front `_blockedTrackIds.removeAll` entirely and let `_markPlaying()` clear it on success.

packages/stream_video/lib/src/webrtc/rtc_audio_api/rtc_audio_html.dart
```diff
@@ -77,3 +235,221 @@ web.HTMLDivElement findOrCreateAudioContainer() {
...
+  }
+
+  String? _computeSkipReason() {
+    if (_disposed) return 'disposed';
+    if (element.srcObject == null) return 'noSrc';
+    if (element.ended) return 'ended';
+    if (_blockedTrackIds.contains(trackId)) return 'blocked';
+    if (!element.paused) return 'notPaused';
+    return null;
+  }
```
renefloor, line 458: Dropping the `readyState >= 2` guard was the right call — it was the thing I couldn't settle from the code last round. But it widens what counts as an outage, so it's worth one real-browser check before merge.

A `suspend` fired shortly after `srcObject` is assigned lands with `paused == true`, `srcObject != null`, `ended == false`, not blocked — so `_computeSkipReason` returns null, `_handlePauseOrSuspend` traces `mediaPlayback.paused`, and a redundant `play()` gets armed 250ms later on *every* track start, followed by a `recover.success` when it lands.

If that's what happens, the new dashboard signal is mostly startup noise. Not something I can settle from the code.

packages/stream_video/lib/src/call/stats/trace_tag.dart
renefloor: Still `navigator.mediaDevices.setSinkId`. The thread got resolved but the string didn't change.

It also now contradicts the comment added directly below at L116-118, which explains the new tags carry no `navigator.mediaDevices.` prefix *because* they aren't device operations. `setSinkId` is `HTMLMediaElement.setSinkId` — same reasoning applies. Worth fixing in this pass; these strings are what dashboards join on, and a wrong name is expensive to correct once data has accumulated under it.

packages/stream_video/lib/src/webrtc/rtc_track/rtc_remote_track.dart
```diff
@@ -66,6 +92,7 @@ class RtcRemoteTrack extends RtcTrack {
       mediaStream: mediaStream ?? this.mediaStream,
       mediaTrack: mediaTrack ?? this.mediaTrack,
       videoDimension: videoDimension ?? this.videoDimension,
+      transceiver: transceiver ?? this.transceiver,
```
renefloor, line 95: The fix itself is right, but it un-noops a call that looks wrong, and the changelog frames restoring that call as the point.

`RtcRemoteTrack.transceiver` comes from `subscriber.onTrack` (`rtc_manager.dart:71`, constructed at `:302`). `unpublishTrack` then passes its sender to `publisher?.pc.removeTrack` (`rtc_manager.dart:353-364`) — a *subscriber* transceiver's sender handed to the *publisher* peer connection.

That branch was dead on any track that had been through `copyWith`; now it's live on every remote-track teardown (call end via `dispose`, participant leave via `removeSubscriber`), on all platforms. Best case it throws into the existing catch and logs a warning; either way it isn't doing what the changelog entry says it does.

Worth deciding whether that branch should exist at all before flipping it on — a remote subscriber track isn't something you `removeTrack` from the publisher. And a one-line `copyWith` test would at least pin the field so this can't silently regress again.

packages/stream_video/lib/src/webrtc/rtc_media_device/rtc_media_device_notifier.dart
```diff
@@ -296,6 +300,27 @@ class RtcMediaDeviceNotifier {
...
+
+  /// Retries playback of the remote audio elements the browser's autoplay
+  /// policy blocked. Must be called from within a user gesture (e.g. a button
+  /// tap) for the browser to allow it.
+  ///
+  /// Observe [CallState.isWebAudioPlaybackBlocked] to know when it is needed.
+  ///
+  /// Web only; a no-op on other platforms. Unrelated to [resumeAudioPlayout],
+  /// which unmutes playout paused via [pauseAudioPlayout].
+  Future<void> resumeWebAudioPlayback() => rtc_audio.resumeAudioPlayback();
```
renefloor, line 324: **API asymmetry, worth settling before this is public.**

You read the condition from `call.state.isWebAudioPlaybackBlocked` and act on it through a global singleton. Every other paired read/act on this surface lives on `Call` — `setAudioOutputDevice`, `setCameraEnabled`, and so on. A three-line `Call.resumeWebAudioPlayback()` forwarder is far cheaper now than after it ships.

Related, and not inline anywhere because it's an absence: nothing in `stream_video_flutter` or the samples surfaces the new flag. The PR introduces a required web affordance and leaves every integrator to build it from the changelog text. Putting it in the dogfooding sample also buys you the manual verification for the whole autoplay path, which no unit test can reach.

packages/stream_video/test/src/call/call_audio_playback_state_test.dart
```diff
@@ -0,0 +1,99 @@
...
+      final call = await joinTestCall();
+
+      deviceNotifier.blocked.add(true);
+      await expectBlocked(call, isBlocked: true);
+
+      deviceNotifier.blocked.add(false);
+      await expectBlocked(call, isBlocked: false);
+    });
+  });
+}
```
renefloor, line 99: These four cover the `Call` <-> notifier wiring well, including the BehaviorSubject seeding rationale.

What's still uncovered is everything with actual logic in it. The watchdog state machine needs a browser, fair enough — but two of the PR's fixes are plain Dart and testable as-is:

- `RtcManager.setAudioOutputDevice`'s new branching (all tracks fail -> `Result.failure`; partial failure -> success; no remote audio tracks -> success). That's the behaviour the changelog promises, and it's three tests over a map of fakes.
- `RtcRemoteTrack.copyWith` preserving `transceiver` — one assertion.

renefloor (approved):

PR #1308, opened 2026-08-18
chore(llc): migrate to TokenManager from core [V2]

Brazol:


renefloor (commented):
_Superseded — I re-posted these findings as inline comments on the relevant lines._

renefloor (commented):
Splitting the earlier review into inline comments — same findings, nits dropped. Three I'd want addressed before merge (core pin, guest token expiry, `peekToken()` mid-refresh); the rest are smaller. Nice cleanup otherwise — good to see the duplicated `TokenManager`/`TokenProvider` go, and the guest profile-fields and anonymous-token fixes are welcome.

pubspec.yaml
```diff
@@ -29,7 +29,7 @@ dependency_overrides:
     git:
       url: https://github.com/GetStream/stream-core-flutter.git
       path: packages/stream_core
-      ref: 31ca84f0885eccea9ad6114719499e6f523991b0
+      ref: dd79170aa48a4f8ab8236a5ad0689167057ac5ec
```
renefloor: **The core pin points at a commit that isn't on `main`.**

`dd79170a` is the *head* of core PR #156, which was squash-merged, so it isn't an ancestor of `stream-core-flutter/main`. The equivalent commit on main is `6f97f0438cf506feed749d02870797e3afc8f0c6` — I diffed `packages/stream_core` between the two and they're byte-identical.

The ref this replaces (`31ca84f0`) *is* on main, so this also breaks the convention here. A PR-head SHA that no branch points at is only fetchable via `refs/pull/*`. Please repoint to `6f97f043`.

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -183,6 +181,57 @@ class StreamVideo extends Disposable {
...
+      // An anonymous token may carry a caller-supplied JWT (e.g. a
+      // call-restricted token granting access to a closed livestream).
+      UserType.anonymous => TokenProvider.static(
+        UserToken.anonymous(userId: user.id, rawValue: userToken ?? ''),
+      ),
+      // The guest loader only dereferences `_client` when a token is actually
+      // requested, after this constructor has assigned it.
+      UserType.guest => TokenProvider.dynamic((userId) async {
+        final existingToken = guestToken;
+        if (existingToken != null) return existingToken;
```
renefloor: **An expired guest token is now permanently unrecoverable.**

`guestToken` is captured in the closure and returned forever, so `refreshToken()` (`expireToken()` + `getToken()`) hands back the *same expired* token. Two consequences:

- `RpcRetryManager` sees `usesStaticProvider == false` for guests, spends its single auth retry on a refresh that changes nothing, then fails.
- `coordinator_ws._authenticateUser` with `_refreshToken: true` re-authenticates with the same dead token → a reconnect loop that can never heal. Previously the guest was re-created and recovered.

I'd also question the premise in the comment (and the changelog entry): the loader is always called with `_tokenManager.userId`, which stays `user.id` — the *original* id, since the server-resolved id only lands in `_state.user.value`. So a second `loadGuest` would send the same id. If the server is deterministic there, re-creating is the safer behaviour.

If it isn't deterministic, I'd still prefer one re-create on expiry over a session that silently dies — or at minimum a terminal, clearly-logged failure instead of a reconnect loop.

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -426,7 +473,7 @@ class StreamVideo extends Disposable {
 
     if (_connectionState.isConnected) {
       _logger.w(() => '[connect] rejected (already connected)');
-      final token = _tokenManager.getCachedToken();
+      final token = _tokenManager.peekToken();
```
renefloor: **`connect()` can now fail while connected, mid-refresh.**

`peekToken()` returns null in the window between `expireToken()` and the new token being cached. The old `TokenManager` kept `_token` populated until the replacement arrived, so this cached read never returned null in the `Connected` state.

A `connect()` racing a WS or 401 refresh now returns `'[connect] userToken is null in Connected state'`. Simplest fix: fall back to `await getTokenAsResult()` when the peek is null.

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -216,37 +265,10 @@ class StreamVideo extends Disposable {
...
-        );
-        return result.data.accessToken;
-      }, onTokenUpdated: onTokenUpdated),
-    };
-
-    _tokenManager.setTokenProvider(user.id, tokenProvider: tokenProvider);
+    // Pre-warm the token cache, mirroring the eager fetch previously
+    // triggered by setting the token provider. Failures surface later,
+    // when the token is actually needed.
+    unawaited(_tokenManager.getTokenAsResult());
```
renefloor: **Token-path error logging was dropped, and this pre-warm discards the result entirely.**

The deleted `_DynamicProvider` logged `[loadToken] failed`, and the old `TokenManager` logged every step. `getTokenAsResult` just converts to `Result.failure`, and this call throws the `Result` away — so a mismatched static token or a throwing `tokenLoader` produces *no* log at construction. That undercuts the "fails fast with an `ArgumentError`" framing in the changelog.

Worth logging on failure inside `getTokenAsResult`, or at least here at the call site.

packages/stream_video/test/src/retry/rpc_retry_manager_test.dart
```diff
@@ -44,13 +43,14 @@ void main() {
 
     setUp(() {
       tokenManager = MockTokenManager();
+      when(() => tokenManager.usesStaticProvider).thenReturn(false);
     });
```
renefloor, line 48: **The new `usesStaticProvider` branch is untested.** Every test in this group stubs it to `false`, so the skip path added in `RpcRetryManager` has no coverage — a test with `thenReturn(true)` asserting `verifyNever(() => tokenManager.getToken())` would pin it.

More broadly, none of the new provider-construction logic in `stream_video.dart` is covered: `_regularTokenProvider`'s initial-token → loader handoff, guest token reuse, anonymous `rawValue` pass-through. `_regularTokenProvider` is already a static method, so `@visibleForTesting` would make it directly testable.

packages/stream_video/test/src/retry/rpc_retry_manager_test.dart
```diff
@@ -44,13 +43,14 @@ void main() {
...
 
     test('refreshes token and retries on 401', () async {
       var callCount = 0;
 
-      when(() => tokenManager.refreshToken()).thenAnswer(
-        (_) async => Result.success(_token),
+      when(() => tokenManager.getToken()).thenAnswer(
+        (_) async => _token,
       );
```
renefloor: These now stub and verify `getToken()` — the *internals* of the `refreshToken()` extension — and never assert that the cache was expired. A `refreshToken()` that forgot to call `expireToken()` would still pass all six tests in this group.

`verifyInOrder([() => tokenManager.expireToken(), () => tokenManager.getToken()])` would pin the actual contract.

packages/stream_video/CHANGELOG.md
```diff
@@ -30,6 +30,18 @@
   - `UserType.authenticated` has been renamed to `UserType.regular`.
   - `UserToken.jwt(String rawValue)` named factory has been replaced by the default `UserToken(String rawValue)` constructor. Replace `UserToken.jwt('...')` with `UserToken('...')`.
   - `UserToken` validation changed: the new constructor throws `ArgumentError` (always) instead of `assert` (debug-only) when the JWT is missing a `user_id` claim.
+- `TokenManager` and `TokenProvider` are now provided by `stream_core`. Two behavioural changes follow:
+  - A static token whose `user_id` claim does not match the connected user's id now fails fast with an `ArgumentError` instead of being rejected later by the server.
+  - `onTokenUpdated` for a static token now fires when the token is first used (and after refreshes) instead of once at client construction; for an initial token combined with a `tokenLoader` it fires at construction.
```
renefloor: **`OnTokenUpdated` losing its `Future` is a behavioural break — and I think the fix belongs in core, not here.**

Core's typedef is `void Function(UserToken)` and `_loadAndNotify` doesn't await it, where the SDK's own typedef was `Future<void> Function(UserToken)` and *was* awaited before the token was returned to the caller. Async callbacks still compile (`void` is a top type for return types), so this passes silently — but an app that persists the token before the SDK uses it loses that ordering.

Rather than documenting it as a break, I'd rather change core's `OnTokenUpdated` back to `Future<void> Function(UserToken)` and await it in `_loadAndNotify`. It's already inside the `synchronized` block, so awaiting keeps the "token is persisted before anyone can use it" guarantee. If we do keep the sync version, this needs its own Breaking entry.

packages/stream_video/CHANGELOG.md
```diff
@@ -30,6 +30,18 @@
   - `UserType.authenticated` has been renamed to `UserType.regular`.
   - `UserToken.jwt(String rawValue)` named factory has been replaced by the default `UserToken(String rawValue)` constructor. Replace `UserToken.jwt('...')` with `UserToken('...')`.
   - `UserToken` validation changed: the new constructor throws `ArgumentError` (always) instead of `assert` (debug-only) when the JWT is missing a `user_id` claim.
+- `TokenManager` and `TokenProvider` are now provided by `stream_core`. Two behavioural changes follow:
```
renefloor: A few more public-API changes worth listing for migrators: `TokenProvider.from` is gone, `TokenProvider.dynamic`'s loader now returns `UserToken` instead of `String`, and `isStatic` has become `usesStaticProvider`.

packages/stream_video/lib/src/token/token.dart
```diff
@@ -13,122 +12,32 @@ typedef GuestTokenLoader = Future<String> Function(UserInfo user);
...
-    OnTokenUpdated? onTokenUpdated,
-  }) {
-    return _DynamicProvider(
-      loader,
-      initialToken: initialToken,
-      onTokenUpdated: onTokenUpdated,
-    );
+/// Result-based convenience wrappers around [TokenManager], bridging its
+/// throwing API to the [Result] flow used across the Video SDK.
+extension TokenManagerResultExtension on TokenManager {
```
renefloor: **This extension is public API.** `src/token/token.dart` is in the barrel (`stream_video.dart`), so `refreshToken`/`getTokenAsResult`/`refreshTokenAsResult` are now attached to core's `TokenManager` for every consumer — whereas the old `TokenManager` was `@internal`. Suggest `@internal`, or moving this to a file that isn't exported.

Also worth noting: core PR #156 explicitly *removed* a `refreshToken` method. If it ever comes back, the instance member will silently shadow this extension.

packages/stream_video/lib/src/stream_video.dart
Brazol: Yes, this was intended change as it aligns with what React SDK is doing with guest users. Previously we were calling createGuest which on the backend side creates new guest id. This will break the token mid session. React makes a guest token static

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -76,6 +75,15 @@ const _idAppState = 2;
...
 const _defaultCoordinatorWsUrl = 'wss://video.stream-io-api.com/video/connect';
 
+/// Creates a server-side guest user, returning its data and access token.
+typedef GuestUserCreator =
+    Future<Result<GuestCreatedData>> Function({
+      required String id,
+      String? name,
+      String? image,
+      required Map<String, Object> custom,
+    });
```
renefloor: If this is a public typedef it's better to have a class for the function data, in case we ever want to add something. Otherwise maybe better to make `@internal` so we don't expose by accident.

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -280,6 +278,89 @@ class StreamVideo extends Disposable {
...
+    UserToken? guestToken;
+
+    return switch (user.type) {
+      UserType.regular => _regularTokenProvider(userToken, tokenLoader),
+      // An anonymous token may carry a caller-supplied JWT (e.g. a
+      // call-restricted token granting access to a closed livestream).
+      UserType.anonymous => TokenProvider.static(
+        UserToken.anonymous(userId: user.id, rawValue: userToken ?? ''),
+      ),
+      UserType.guest => TokenProvider.dynamic((userId) async {
```
renefloor: **Re-entrant `getToken()` deadlocks the client permanently.**

This loader runs *inside* core's `synchronized` block, and the HTTP call it makes can re-enter that same lock:

`TokenManager.getToken()` → `synchronized` → this loader → `createGuest` → `_client.loadGuest` → `CoordinatorClientRetry` → `RpcRetryManager.execute` → 401 → `usesStaticProvider` is **false** (a guest is a `DynamicTokenProvider`) → `refreshTokenAsResult()` → `expireToken(); getToken()` → `synchronized` on the same instance.

`objectSynchronized` uses `BasicLock`, which is non-reentrant, and no `timeout` is passed. Running the exact `getToken`/`_loadAndNotify`/`refreshToken` shape against `synchronized: ^3.4.1`:

```
calling getToken() ...
  loader: creating guest guest-1 ...
  loader: createGuest returned 401 -> RpcRetryManager refreshes token
result: TIMED OUT -> deadlocked
```

The constructor's pre-warm and every later `connect()` join the same lock, so the client hangs with no error and no timeout.

This is latent on `v2` as well — the old `_tokenOperation` memo made the nested call await itself — but this PR rewrites exactly this path, and the fix is cheap: `loadGuest` builds its own `Dio` authenticated with `UserToken.anonymous(userId: id)` and never touches the managed token, so a 401 there should not trigger a *user token* refresh at all. Either hand `loadGuest` an `RpcRetryManager` without a `tokenManager`, or make `refreshToken` re-entrance-safe.

packages/stream_video/lib/src/coordinator/open_api/coordinator_ws.dart
```diff
@@ -113,7 +113,9 @@ class CoordinatorWebSocket {
 
   Future<void> _authenticateUser() async {
     _logger.i(() => '[authenticateUser] url: ${_client.options.url}');
-    final tokenResult = await tokenManager.getToken(refresh: _refreshToken);
+    final tokenResult = _refreshToken
+        ? await tokenManager.refreshTokenAsResult()
```
renefloor: **`usesStaticProvider` guards the HTTP path but not this one.**

`RpcRetryManager` now skips the refresh for static providers, on the grounds that they can only hand back the same token. This call still refreshes unconditionally.

For a regular user constructed with only `userToken`, `_refreshToken` is set by `onTokenError` and never cleared, so every subsequent reconnect runs `expireToken()` + `loadToken()` and gets the same dead token back — the WS reconnects forever with credentials that cannot change. Guests hit the same thing now that the token is memoized for the client's lifetime, even though `usesStaticProvider` reports `false` for them.

Mirroring the guard here would keep the two paths consistent.

renefloor (commented):
Two follow-ups on the token path, plus one still-open item on the core pin.

pubspec.yaml
renefloor: Still open on the latest revision. `gh api repos/GetStream/stream-core-flutter/compare/main...dd79170aa48a4f8ab8236a5ad0689167057ac5ec` returns `"diverged"`, so the pin is still not reachable from core's `main` and is only fetchable via `refs/pull/*`.

`6f97f0438cf506feed749d02870797e3afc8f0c6` remains the equivalent commit on `main`.

packages/stream_video/lib/src/coordinator/open_api/coordinator_ws.dart
```diff
@@ -113,7 +113,12 @@ class CoordinatorWebSocket {
 
   Future<void> _authenticateUser() async {
     _logger.i(() => '[authenticateUser] url: ${_client.options.url}');
-    final tokenResult = await tokenManager.getToken(refresh: _refreshToken);
+    // Mirrors the RpcRetryManager guard: a static provider can only return
+    // the same token again, so refreshing it is pointless.
+    final shouldRefresh = _refreshToken && !tokenManager.usesStaticProvider;
```
renefloor: **A memoized guest token has exactly the property this guard is testing for, but fails the predicate.**

The comment is right that "a static provider can only return the same token again" — and since #1308 the guest provider memoizes `guestToken` for the client's lifetime, so it can only return the same token again too. But it's a `DynamicTokenProvider`, so `usesStaticProvider` is `false` and it takes the refresh branch.

Concretely, for a guest whose token the server rejects:

`onTokenError` → `_refreshToken = true` → `refreshTokenAsResult()` → `expireToken()` + the memoized token → rejected again → `onTokenError` → … and `_refreshToken` only clears on `CoordinatorConnectedEvent` (line 153), which never arrives. The same applies to the `RpcRetryManager` guard, which burns its one auth retry on a refresh that cannot change anything.

Following on from the guest-token thread: taking the React-parity behaviour as settled, the cheap way to make both guards correct is to let the provider describe itself accurately — swap to `TokenProvider.static(guestToken)` after the first successful load, using core's `tokenProvider` setter. Then `usesStaticProvider` becomes `true` for guests and this line, plus `RpcRetryManager`, both do the right thing with no extra predicates.

Failing that, a guest token rejection should end as a terminal, clearly-logged failure rather than an unbounded reconnect loop.

renefloor (commented):
One follow-up on the new refresh guard.

renefloor (approved):

packages/stream_video/lib/src/coordinator/open_api/coordinator_ws.dart
```diff
@@ -180,16 +207,20 @@ class CoordinatorWebSocket {
     final source = state.source;
     final wsException = source is ServerInitiated ? source.error : null;
 
+    final apiError = wsException?.apiError;
```
renefloor: **The new terminal auth failure reaches the app with no reason attached.**

This only unwraps an error from `ServerInitiated`, but the failure mode this PR adds arrives as `AuthenticationFailed`. Core wires the authenticator's throw straight to it:

```dart
// stream_web_socket_client.dart
onFailure: (error) => disconnect(source: .authenticationFailed(error: error)),
```

So when `_authenticateUser` throws — the static-provider guard above at line 127, a `tokenLoader` that threw, or `send` failing — `source` is `AuthenticationFailed`, `wsException` is `null`, and the emitted `CoordinatorDisconnectedEvent` carries `closeCode: null`, `closeReason: null`, `apiError: null`. The `VideoError` with the actual explanation is dropped, and the app can't tell a refused token from any other disconnect.

That's the opposite of what the changelog promises for this path:

> A coordinator WebSocket connection whose token could not be loaded is now closed as an authentication failure instead of being retried indefinitely. […] the error is reported on `StreamCallDisconnectedEvent`/`CoordinatorDisconnected` — where before the socket kept retrying with nothing reported.

Ending the connection instead of looping is the important half and it works. But it's now terminal *and* silent, so "nothing reported" still holds — and since `isReconnectable` is `false` for `AuthenticationFailed`, this event is the app's only notification.

`sfu_ws.dart` in this same PR does handle it (`case AuthenticationFailed(): … source.error`), which is what makes me read this as an oversight rather than a decision.

Core's `Disconnected.cause` already flattens both sources:

```dart
Object? get cause => switch (this) {
  ServerInitiated(:final error) => error?.error ?? error,
  AuthenticationFailed(:final error) => error,
  …
};
```

So either read `source.cause` here, or add the branch explicitly and fall `closeReason` back to it — whichever you prefer, as long as an `AuthenticationFailed` disconnect says why. A test on this handler would be worth having alongside it, given it's the only account of a terminal auth failure.

renefloor (commented):
Re-reviewed the five commits since my approval. The `TokenSource`/`buildTokenProvider`/`establishGuestSession` split is a genuine improvement over what I reviewed before, and it closes every substantive thread I had open:

- **Core pin** — `f83b5d4d` is an ancestor of core's `main` now (`compare` reports `behind`, not `diverged`). Resolved.
- **Re-entrant `getToken()` deadlock** — `loadGuest` runs through the new `_unauthenticatedRetryManager`, so guest creation can no longer trigger a user-token refresh. Core also dropped `synchronized` for a generation-keyed in-flight cache, so the `expireToken(); getToken()` shape is safe on its own. Resolved, and the load-bearing reason is documented at the field.
- **Guest token / `usesStaticProvider` mismatch** — the guest installs `TokenProvider.static` after establishment, so both refresh guards now classify guests correctly, and `TokenSource` keeps the pre-identity window honest. This is the shape I suggested; nice.
- **`peekToken()` mid-refresh** — `connect()` goes through `_tokens.getToken()`. Resolved.
- **Unbounded reconnect on a dead token** — the WS guard mirrors the RPC guard, and codes 41–43/2 are non-reconnectable in core while 40 gets exactly one refresh. Both branches terminate.
- **Extension as public API / internal typedefs** — `token_manager_extension.dart`, `token_provider_factory.dart`, and `token_source.dart` are all out of the barrel. Resolved.
- **Missing token-path logging** — `getTokenAsResult`/`refreshTokenAsResult` log on failure. Resolved.
- **Test coverage** — `build_token_provider_test.dart` and `token_source_test.dart` cover the provider combinations, guest establishment, and the concurrency/non-sticky-failure behaviour, and the static-provider skip branch is pinned. Resolved.
- **`OnTokenUpdated` losing its `Future`** — you kept core's sync signature and gave it a Breaking entry, which was the alternative I offered. Fine by me.

One new finding, on the auth-failure path this PR introduces — inline. It's small, and it's the one thing I'd fix before merge because it undercuts the changelog entry that goes with it.

Not blocking, just noting: nothing covers `CoordinatorWebSocket._onConnectionStateChanged`, which is why that gap went unseen — the codec side of the same feature is well tested.

renefloor (approved):

PR #1310, opened 2026-08-20
feat(ui, samples): introduced componentBuilders with participant tile

renefloor:
### 🎯 Goal

Introducing the component builders from core and chat.

### 🛠 Implementation details

This creates a `streamVideoComponentBuilders` in which we can add more component builders later.
It creates a new `StreamParticipantTile` and renamed the old `StreamCallParticipant` to `DefaultStreamParticipantTile`. The new naming matches figma: https://www.figma.com/design/BJppI3JYlPy3XeEu5MZY1E/SDK-Design-System?node-id=24212-75268&p=f&m=dev

To keep migrations easy and simple we kept the old `StreamCallParticipant`, but deprecated it and added a dart fix rule.

The dogfooding app adds a no-op component factory as demo:
```dart
  late final _componentBuilders = StreamComponentBuilders(
    extensions: [
      ...streamVideoComponentBuilders(
        // No-op change for [StreamParticipantTile] as demo example.
        participantTile: (context, props) =>
            DefaultStreamParticipantTile(props: props),
      ),
      // You can combine both chat and video component builders.
      ...streamChatComponentBuilders(),
    ],
  );
```

### 🎨 UI Changes

Should be no UI changes.

## Summary by CodeRabbit

* **New Features**
  * Added configurable participant tiles with app-wide component customization.
  * Added support for custom video renderers, placeholders, labels, styling, reactions, and connection indicators.
  * Added shared Stream Video and Chat component configuration.

* **Improvements**
  * Updated calls, livestreams, screen sharing, and Android picture-in-picture to use the new participant tile experience.

* **Migration**
  * Replaced `StreamCallParticipant` with `StreamParticipantTile`.
  * `StreamCallParticipant` remains available as a deprecated compatibility option; use `dart fix --apply` to migrate automatically.


packages/stream_video_flutter/lib/src/call_participants/participant_tile.dart
```diff
@@ -0,0 +1,520 @@
...
+
+/// A widget that represents a single participant in a call.
+///
+/// Kept as a thin wrapper around [DefaultStreamParticipantTile] so existing
+/// code keeps working; it takes the same parameters as before.
+@Deprecated(
+  'Use StreamParticipantTile instead. '
+  'Will be removed in the next major version.',
+)
+class StreamCallParticipant extends StatelessWidget {
```
xsahil03x, line 454: I think we can just make it a typedef for the new widget instead.

renefloor: Yeah typedef could also be. We still have to update the widget itself though, so we might want to use this widget to map the legacy theme to the new participant theme if we make changes there. We can still go both ways.

packages/stream_video_flutter/CHANGELOG.md
```diff
@@ -1,5 +1,33 @@
 ## Upcoming (next major)
 
+### ✅ Added
+
+- Added component factory support, so the Stream Video UI components can be replaced app-wide instead of threading widget builders through the widget tree. Register the components you want to replace with `streamVideoComponentBuilders` and wrap your app in a `StreamComponentFactory`:
+
+  ```dart
+  MaterialApp(
+    builder: (context, child) => StreamComponentFactory(
+      builders: StreamComponentBuilders(
```
Brazol: We shouldn't recommend to create it inline right? Following the comment from the usage in dogfooding:

/// Built once so its identity stays stable across rebuilds, otherwise every
/// widget depending on the [StreamComponentFactory] rebuilds on each frame.
late final _componentBuilders = StreamComponentBuilders(...);

renefloor: I think it depends a bit. If you just have a simple MaterialApp it shouldn't really rebuild a lot. Not sure if the doc in the dogfooding app actually makes sense. Let me check why that would be rebuilding.

renefloor: Checked it, and you're right that the identity matters — though for a different reason than my comment in the dogfooding app claims.

`StreamComponentFactory.updateShouldNotify` is `builders != oldWidget.builders`, and the generated `==` compares `extensions` with `Map ==`, which is identity. The constructor always builds a fresh map, so two freshly built instances are never equal — even `StreamComponentBuilders() == StreamComponentBuilders()` is `false`. Deep map equality wouldn't rescue it either, since inline closures are never `==` across rebuilds. So constructing it inline always invalidates every widget that resolved a builder through the factory.

What is *not* true is "rebuilds on each frame". `MaterialApp.builder` sits above the Router/Navigator (`WidgetsApp` wraps `routing` in a `Builder`), so it only re-runs when `WidgetsApp` itself rebuilds. Measured in a widget test: idle frames, window resize/rotation and keyboard insets do not re-run it; a platform brightness change or a `setState` above does. And when that happens `routing` is recreated too, so the subtree rebuilds anyway.

So inline inside `MaterialApp.builder` is close to harmless, but the recommendation stands: it costs nothing, and the factory won't always sit at the app root — inside a screen that rebuilds (animation, `StreamBuilder`, `setState`) it would rebuild every Stream component. I've hoisted the builders in the CHANGELOG snippet and reworded the dogfooding comment to say what actually happens.

renefloor: fixed: https://github.com/GetStream/stream-video-flutter/pull/1310/commits/a2ac6b4e737c6fb461a51d14ca851aaa753719d0

Brazol (approved):

PR #1312, opened 2026-08-21
feat(llc): end-to-end encryption

Brazol:
Part of FLU-171
Flutter Webrtc part - https://github.com/GetStream/webrtc-flutter/pull/83
Docs - https://github.com/GetStream/docs-content/pull/1537

Adds end-to-end encryption to the Flutter SDK, backed by the native `EncryptionManager` in the `flutter-webrtc` (PR #110).

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end encryption for calls with shared-key and passphrase-based setup.
  * Added encryption-key support for join links and call sharing.
  * Added in-call encryption indicators, key mismatch alerts, and key replacement.
  * Added encryption status reporting and public SDK configuration APIs.
* **Bug Fixes**
  * Improved handling of missing calls and encryption-key resolution failures.
  * Prevented media publication when encryption setup fails, avoiding unencrypted transmission.
  * Stopped retries for errors marked unrecoverable by the server.


renefloor (commented):
### Automated review pass — E2EE

Reviewed at `69881cbd`. Five findings below that each end in either cleartext media or a wrong encryption indicator, so I'd treat them as merge blockers rather than follow-ups. Every one was verified against the source.

**Also worth knowing (not inline):**

- **Two mutation-proven test gaps.** Deleting `unawaited(attachDecryptor(...))` *and* both `flushPendingDecryptors()` call sites leaves all 518 tests green. Same for removing both `sfuE2eeEnabledUpdated()` call sites. Both features can be deleted without a single failure.
- **The resolver happy path is untestable.** `EncryptionManager.create` is a static with no injection seam, so every resolver test returning a non-null key necessarily ends in `Failure`. `await manager.setSharedKey(keyIndex, bytes)` never executes in any test — swapping those two arguments would ship green, and `keyIndex` is the one value whose mismatch fails every decrypt.
- **`props` excludes `bytes`** (`call_encryption_key.dart`). Two keys at the same index with different material compare equal, so key rotation guarded by `if (newKey != current)` silently never re-applies. The stated reason doesn't hold — `toString()` is overridden directly below, so the props-based stringify was already unreachable.
- **Platform check ordered after the resolver** (`call.dart:1148`). Setting `encryptionKeyResolver` on `StreamVideoOptions.defaultCallPreferences` — which the docs recommend — fails *every* join on web/Windows/Linux, including `available`-mode calls that would join fine unencrypted.
- **Docs:** `call.dart:1037` has `await call.join(create: true)`, but `join()` has no `create` parameter. `call.dart:1116` tells developers to set `encryptionKeyResolver` on `StreamVideoOptions`; the field is `StreamVideoOptions.defaultCallPreferences`.
- **`StreamEncryptionMode.fromString`'s `default: disabled` is dead code** — `encryption_settings_response.dart:62` does `fromJson(...)!`, which throws on an unknown value first. And since `required` is derived from the mode, defaulting unknown to `disabled` is fail-*open* for a security setting.
- **dogfooding:** `random_words.dart:113` uses `Random()`, not `Random.secure()`, for the passphrase — while `e2ee.dart` correctly uses `Random.secure()` for raw keys. The ~20-bit entropy is honestly documented as demo-only, but this is code customers copy.

**Credit where due:** the encryptor side is genuinely fail-closed — one choke point for every sender, attach before negotiation, and a failed attach stops the transceiver instead of publishing cleartext, with a comment explaining why softening it would be wrong. Reconnect/rejoin/migration re-attachment is correct by construction, and `clearE2EEManager` in `leave()` does not break reconnect. The CallKit-watchdog latency note and the surplus-manager race explanation are load-bearing comments a maintainer couldn't derive from the call site.

packages/stream_video/lib/src/call/call.dart
```diff
@@ -1000,6 +1017,243 @@ class Call {
...
+  /// await e2ee.setSharedKey(0, sharedKeyBytes);
+  ///
+  /// await call.setE2EEManager(e2ee);
+  /// await call.join(create: true);
+  /// ```
+  ///
+  /// The manager is released and disposed when the call is left, so a fresh
+  /// one is needed per call.
+  Future<void> setE2EEManager(EncryptionManager manager) async {
+    if (_session?.rtcManager != null) {
```
renefloor: **Critical — a manager attached mid-join is accepted, and the session publishes cleartext anyway.**

This guard tests `rtcManager`, but `rtcManager` is only assigned inside `CallSession.start()` *after* the SFU join response arrives. The session captured the manager by value much earlier, at `call.dart:1591` (`e2eeManager: _e2eeManager`).

So for the whole join window — WS connect, `joinCall` round trip, SFU join response — `_session?.rtcManager` is `null` and this guard passes.

Failure scenario:
1. `call.join()` is started without awaiting. The SDK itself does this: `unawaited(callToJoin.join())` at `stream_video.dart:989`.
2. `_resolveE2EEManager()` finds no manager, `joinCall` goes out with `e2ee: false`, session built with `e2eeManager: null`.
3. App calls `await call.setE2EEManager(m)`. No throw. `call.e2eeManager` now returns `m`.
4. `RtcManager` was built with `null`, so `_attachEncryptor` is a success-no-op and every local track publishes in cleartext — while the app believes the call is encrypted.

No error and no log at any point. The error message here also names "peer connections" as the reason, which points at the wrong cutoff.

Suggestion: gate on the join having started rather than on peer connections existing — `_session != null`, or a status check (`lifecycleCallConnecting` is set early at `call.dart:1532`), or both.

packages/stream_video/lib/src/call/call.dart
```diff
@@ -1000,6 +1017,243 @@ class Call {
...
+    );
+  }
+
+  /// Builds and attaches an [EncryptionManager] from the app's key resolver,
+  /// for a call that does not have one yet.
+  Future<Result<None>> _resolveE2EEManager() async {
+    // An explicitly attached manager wins.
+    if (_e2eeManager != null) return const Result.success(none);
+
+    final required =
```
renefloor: **Critical — the encryption mode is read before the coordinator has been asked for it.**

`_resolveE2EEManager()` is called at `call.dart:1317`, i.e. before `_join` → `_doJoin` → `joinCall`, which is where settings first reach state. `_init()` fetches nothing, and `CallSettings.encryption` defaults to `StreamEncryptionMode.disabled`.

So for `makeCall(...)` followed straight by `join()` — the exact pattern in `setE2EEManager`'s own docstring — this reads `disabled` regardless of the call's real mode. Two consequences:

1. `required` is always `false`, so the documented "This call requires end-to-end encryption, but no key was available" rejection never fires for an `auto-on` call on that path.
2. The resolver is handed `encryptionMode: disabled` — actively wrong, not merely unknown. `CallEncryptionKeyRequest.encryptionMode` is documented as being there "so a resolver can answer only for the calls that need it", and this PR's own sample does exactly that: `ringing_encryption.dart:10` returns `null` unless the mode is `autoOn`. Result: the join goes out with `e2ee: false` and the server rejects it, so the app sees an opaque join failure instead of an encrypted call.

The ringing path happens to work (`consumeIncomingCall` does a `getCall` first) and the dogfooding lobby happens to work (`lookupCallExists` calls `call.get()`), but neither is required of an SDK consumer.

Note `call_e2ee_resolver_test.dart:96` masks this — `_stateWith(mode: ...)` pre-seeds `state.settings` via `copyWith`, so the test never exercises the real ordering. A test that only stubs `getOrCreateCall`/`joinCall` with `autoOn` settings and then calls `join()` would fail today.

Suggestion: resolve after the join response returns metadata but before the session starts, or `get()` first when a resolver is configured. Also consider letting `CallEncryptionKeyRequest` express "mode not yet known" so resolvers don't answer confidently on default data.

packages/stream_video/lib/src/call/call.dart
```diff
@@ -1000,6 +1017,243 @@ class Call {
...
+    _e2eeManager = null;
+
+    if (identical(_e2eeClaims[callCid.value]?.call.target, this)) {
+      _e2eeClaims.remove(callCid.value);
+    }
+
+    await _e2eeEventsSubscription?.cancel();
+    _e2eeEventsSubscription = null;
+
+    if (dispose) {
```
renefloor: **Critical — this disposes a manager another live `Call` may still be using.**

`setE2EEManager` deliberately permits two `Call` instances for the same cid to share one manager, and `call_e2ee_test.dart:357` asserts that as intended behaviour ("the same manager on another instance of the call is allowed").

But this dispose is unconditional. The claimant check above only guards *claim removal* (`if (identical(_e2eeClaims[callCid.value]?.call.target, this))`), not disposal. So `first.clearE2EEManager()` — or `first.leave()`, which calls this with `dispose: true` at `call.dart:2783` — tears down the native key store that `second` is actively encrypting and decrypting with. Every later frame on the live call fails and the user watches a working call degrade.

Duplicate same-cid `Call` instances are normal here (the ringing/push paths, which is precisely what `encryptionKeyResolver` was added for).

Related, same method: the warning above is gated on `dispose && _session?.rtcManager != null`, so the `dispose: false` hand-off the docstring actively recommends gets no warning at all. `RtcManager` and `CallSession` hold their own `final EncryptionManager?`, so clearing the `Call` field detaches nothing — call A keeps using the manager while call B legitimately claims it. That is "two calls, one key store", which is the exact thing `_e2eeClaims` exists to prevent; the registry guards *attached* managers, not *in-use* ones.

Suggestion: skip disposal when this call isn't the claimant, and reject (or warn loudly on) any release while `_session?.rtcManager != null`, dispose or not.

packages/stream_video/lib/src/call/state/mixins/state_sfu_mixin.dart
```diff
@@ -173,6 +173,11 @@ mixin StateSfuMixin on StateNotifier<CallState>, StatePendingTracksMixin {
     );
   }
 
+  /// Records whether the SFU considers this call end-to-end encrypted.
+  void sfuE2eeEnabledUpdated(bool isE2eeEnabled) {
```
renefloor, line 177: **Critical — `isE2eeEnabled` is the SFU's opinion, and nothing reconciles it with whether this client actually encrypts.**

This records a server flag and stops. Nothing compares it against `_e2eeManager != null`, and nothing knows whether the local transforms attached. Yet the CHANGELOG tells apps to read `CallState.isE2eeEnabled` "to check whether it is in effect", and dogfooding renders a shield with "This call is end-to-end encrypted" straight off this boolean.

Two silent divergences:
- **SFU says encrypted, we have no manager.** Reachable via the mid-join race (`call.dart:1043`), via a rejoin after `leave()` disposed the manager, or an `available`-mode call where the resolver returned `null`. We publish cleartext; the badge says encrypted.
- **Manager attached but the decryptor never attached** (`rtc_manager.dart:665`). Badge says encrypted; participant tiles are permanently black.

Separately, it is never reset on teardown: `lifecycleCallEnded` and `lifecycleCallDisconnected` both clear `sessionId` and `callParticipants` but not this flag. After leaving an encrypted call, state reads `isE2eeEnabled: true` with `e2eeManager == null` — two sources of truth that contradict each other, neither authoritative.

Suggestions:
1. Reset `isE2eeEnabled: false` in `lifecycleCallDisconnected` and `lifecycleCallEnded` (one line each).
2. Consider renaming to what it means (`isSfuE2eeNegotiated`) and adding a derived signal for the question apps actually ask, combining manager presence with the SFU flag and last-known decryption health.
3. `_onE2eeEvent` (`call.dart:1239`) already discriminates `missingKey` / `decryptionFailed` / `encryptionFailed` / `unencryptedFrame` by severity and routes all of it to the logger. Feeding it into state as a coarse `off`/`active`/`failing` enum turns "the padlock lies" into "the padlock can go amber" — highest-value change here.

packages/stream_video/lib/src/webrtc/rtc_manager.dart
```diff
@@ -638,6 +653,89 @@ class RtcManager extends Disposable {
...
+  /// The user id that owns the remote track under [trackIdPrefix].
+  String? _userIdForTrackPrefix(String trackIdPrefix) {
+    return stateManager.callState.callParticipants
+        .firstWhereOrNull((it) => it.trackIdPrefix == trackIdPrefix)
+        ?.userId;
+  }
+
+  /// Decrypts everything [track]'s receiver delivers, or queues the track
+  /// until its participant is known.
+  Future<void> attachDecryptor(RtcRemoteTrack track) async {
```
renefloor, line 665: **Critical — a failed decryptor attach is invisible to the app *and* to the E2EE event stream.**

Three exits install no transform: `userId == null` (queued, no timeout, no terminal failure), `receiver == null` (`_logger.w`, return), and `manager.decrypt` throwing (`_logger.e`, return).

This is fired as `unawaited(attachDecryptor(remoteTrack))` at `rtc_manager.dart:317`, and the track is registered regardless — `onRemoteTrackReceived?.call(...)` and `tracks[...] = ...` run on the next two lines, and `_onTrackPublished` may already have called `track.start()`. So the app gets a normal-looking participant whose video is black and audio silent, forever.

What makes this worse than ordinary log-and-continue: **because no transform was installed, native is never in the loop for that track, so it can never emit `decryptionStalled`, `missingKey` or `decryptionFailed` for it.** Every app-side detection strategy this PR ships is structurally blind to it — including the dogfooding banner, which keys entirely off `decryptionStalled` (`e2ee_key_notification.dart:109`). The only diagnostic is an SDK log line.

Aggravating details:
- `flushPendingDecryptors` is driven only by `SfuJoinResponseEvent` and `SfuParticipantJoinedEvent`. `sfuParticipantUpdated` can be what first makes a `trackIdPrefix` resolvable and does not flush.
- `_pendingDecryptors` is never pruned when a participant leaves, so departed participants' tracks are retained for the `RtcManager`'s lifetime and retried on every flush.
- `flushPendingDecryptors` clears the map before awaiting each attach, so a track queued concurrently during the flush isn't picked up by it.

Also untested — deleting the `unawaited(attachDecryptor(...))` call and both `flushPendingDecryptors()` call sites leaves all 518 tests green.

Suggestion: return a `Result` so `_onRemoteTrack` can record the outcome, and surface a per-track "encrypted, no decryptor installed" condition (call-level event or participant flag) with a bounded retry before it becomes terminal.

dogfooding/lib/screens/lobby_screen.dart
```diff
@@ -10,19 +10,49 @@ import 'package:stream_video_flutter/stream_video_flutter.dart';
...
+      required StreamVideoEffectsManager effectsManager,
+      String? encryptionKey,
+    });
+
 class LobbyScreen extends StatefulWidget {
   const LobbyScreen({
     super.key,
     required this.onJoinCallPressed,
     required this.call,
+    this.callExists = true,
```
renefloor, line 39: Wouldn't it make more sense to default to false? Or maybe make it nullable and do a lookup in the initState when it's null?

I can also make such changes in the v2 branch to make it more compatible with a generic lobby screen and new path system.

Brazol: yes might be a good idea to do them together with changes in v2

renefloor (commented):
### Follow-up: gaps missing from my earlier pass

Completing [my earlier review](https://github.com/GetStream/stream-video-flutter/pull/1312#pullrequestreview-5103534978) with four findings it didn't cover. Four inline comments below.

The first one came out of the cross-SDK thread on whether a failed encryptor attach should publish cleartext or refuse and tell the app. **Flutter gets the "refuse" half right, and does it more robustly than bubbling the exception up — but the "tell the app" half is missing on the join path.**

#### Verified non-issue, recorded so nobody re-chases it

I traced the negotiation race and it is **not** a bug — worth writing down because it looks like one:

`pc.addTransceiver` fires `negotiationneeded` natively, and `onRenegotiationNeeded → negotiateOrRecover` is fire-and-forget (`onRenegotiationNeeded?.call(this)` is not awaited). So an offer genuinely can be in flight while `_attachEncryptor` is still awaiting the platform channel. It is closed by the pending-transceiver gate:

- `_addTransceiver` registers `_pendingTransceivers[key]` **before** calling `_createTransceiver` and clears it in `whenComplete`, so the claim spans the whole attach *and* the stop.
- `getAnnouncedTracks` — inside the negotiation lock, after `createOffer()` — starts with `await _settlePendingTransceivers()`.

So negotiation blocks until the attach resolves. A failed transceiver is stopped and never entered into `transceiversManager`, so it cannot be announced. If it was the only track, `tracksInfo.isEmpty` → `rollbackLocalDescription()` and **no SetPublisher at all**, matching JS. If other tracks were already publishing, SetPublisher covers only those — arguably better than failing the whole negotiation. The 5s `_pendingTransceiverSettleTimeout` also fails closed (returns `null` → rollback → fast reconnect).

Nice piece of design; it just isn't obvious from reading `_createTransceiver` alone.

packages/stream_video/lib/src/webrtc/rtc_manager.dart
```diff
@@ -1463,6 +1562,46 @@ extension PublisherRtcManager on RtcManager {
...
+    } catch (e, stk) {
+      _logger.e(
+        () =>
+            '[attachEncryptor] failed for trackType: '
+            '${publishOptions.trackType}; refusing to publish cleartext: $e',
+      );
+      _logger.v(() => '[attachEncryptor] $stk');
+
+      return Result.error(
+        'Failed to attach the E2EE encryptor for trackType: '
```
renefloor, line 1649: **Important — this error is correct, and then goes nowhere on the join path.**

Refusing to publish rather than sending cleartext is the right call, and the surrounding machinery genuinely delivers it (see the non-issue note in the review body). The problem is purely that nobody hears about it.

Two paths diverge:

- **App-initiated publish** — `call.setCameraEnabled(...)` → the `Result.error` propagates to the caller. ✅
- **Join path** — swallowed. `_applyConnectOptions` (`call.dart:2908`) awaits `_applyCameraOption` / `_applyMicrophoneOption` / `_applyScreenShareOption` and **discards their `Result`s** — it is declared `Future<void>`. And it is itself invoked as `_applyConnectOptions().catchError(log)` at `call.dart:2011`, which catches *thrown* errors, not returned `Failure`s.

Failure scenario: the platform transform bridge hiccups during join. The user lands in a call that reports connected — and, per the `isE2eeEnabled` comment on my earlier review, badges as encrypted — with camera and mic silently unpublished and mute state showing disabled. Indistinguishable from "I forgot to unmute."

So for the question raised in the cross-SDK thread: Flutter answers "don't publish at all" correctly, but "inform the app" only holds for explicit publish calls, not for join.

Suggestion: propagate the failure out of `_applyConnectOptions` (return the `Result`s, or surface via call state / an event) so a refusal-to-publish is visible rather than log-only. A deliberate decision to fail closed deserves a user-visible consequence.

Minor, same area: `_stopTransceiver` (`rtc_manager.dart:1707`) swallows a failing `stop()` with a warning while the caller still returns `Result.error`, so nobody learns the sender survived. That's the one remaining path where a live unencrypted sender could linger — still unannounced, so the SFU can't map it, but it's the weak link in an otherwise clean fail-closed chain.

packages/stream_video/lib/src/call/call.dart
```diff
@@ -1000,6 +1017,243 @@ class Call {
...
+    await _e2eeEventsSubscription?.cancel();
+    _e2eeEventsSubscription = manager.events.listen(
+      _onE2eeEvent,
+      onError: (Object e) => _logger.w(() => '[e2ee] event stream error: $e'),
+    );
+  }
+
+  /// Builds and attaches an [EncryptionManager] from the app's key resolver,
+  /// for a call that does not have one yet.
+  Future<Result<None>> _resolveE2EEManager() async {
```
renefloor, line 1109: **Important — `join()` on an already-left `Call` creates and then leaks an `EncryptionManager`, and poisons the claim registry for that cid.**

`_callLifecycleCompleter` is `final` and never reset, so a `Call` is single-use: after a leave, `_doJoin` bails at `call.dart:1527` with `'call was left'`. But this runs earlier — from `join()` at `call.dart:1317` — and never checks the lifecycle.

Sequence on a second `join()` of a left call, with `encryptionKeyResolver` configured:

1. The resolver is invoked; `EncryptionManager.create` + `setSharedKey` run; `setE2EEManager` registers a claim at `call.dart:1093`.
2. `_doJoin` immediately returns `'call was left'`.
3. `join()`'s failure branch calls `leave()` → `_disconnect` hits `if (state.value.status.isDisconnected) return false` at `call.dart:2662`, which is **before** the `try/finally` that calls `_clear`. So `clearE2EEManager()` never runs.
4. Net result: the native manager is never disposed, `_e2eeEventsSubscription` is never cancelled, and the claim stays in the static map keyed to a **live** `Call` target — so the `WeakReference` will never clear and the opportunistic `removeWhere` sweep can never prune it.

That leaked claim then makes the cross-instance guard fail deterministically: every subsequent `makeCall(cid).setE2EEManager(...)` for the same cid throws `StateError` for the remainder of the process.

Suggestion: skip `_resolveE2EEManager` when `_callLifecycleCompleter.isCompleted`, and/or release the manager on that bail-out path.

packages/stream_video/lib/src/call/call.dart
```diff
@@ -1000,6 +1017,243 @@ class Call {
...
+    if (current != null && !identical(current, manager)) {
+      throw StateError(
+        'This call already has a different EncryptionManager. Overwriting it '
+        'would drop the current one while it still holds a native key store, '
+        'so release it first: clearE2EEManager() to dispose it, or '
+        'clearE2EEManager(dispose: false) to hand it to another call.',
+      );
+    }
+
+    final claimant = _e2eeClaims[callCid.value]?.call.target;
```
renefloor, line 1074: **Important — this guard's outcome depends on GC timing.**

`StreamVideo.makeCall` always constructs a fresh `Call`, so two `Call` objects for one cid are routine. The claim is released only in `clearE2EEManager`, reached only via `_clear` (`call.dart:2783`) — i.e. only on `leave()`/disconnect. A `Call` that got a manager in a lobby and was navigated away from without `leave()` keeps its claim until the GC clears the `WeakReference`, and `WeakReference.target` gives no promptness guarantee.

Failure scenario: user opens the encrypted lobby (`callA = makeCall(cid)`, `setE2EEManager(m1)`), backs out without leaving, re-opens it (`callB = makeCall(cid)`, `setE2EEManager(m2)`). Whether this throws depends purely on whether the VM has collected `callA` — identical code, different outcome run to run.

Both branches are bad in different ways: throwing surprises the app for reasons outside its control, and passing installs a second key store on one cid, which is the exact silent one-way-media bug the registry exists to prevent. It's currently fail-open.

Note the passing test only works because the first `Call` is still strongly held by the test; the complementary "a stale claim from a dropped call is ignored" case is untestable by construction.

Suggestion: make staleness lifecycle-driven rather than GC-driven — hold the `Call` side strongly and remove the entry in `clearE2EEManager` — or key the check on something observable like membership in `_streamVideo.state.activeCalls`. Worth reconsidering whether this belongs on `Call`'s statics at all; needing `@visibleForTesting resetE2EEClaims()` is a hint the invariant wants a scoped owner (the `StreamVideo` client), since as written it is also shared across every client in the process.

packages/stream_video/lib/src/call/call.dart
```diff
@@ -1000,6 +1017,243 @@ class Call {
...
+            await manager.setSharedKey(keyIndex, bytes);
+        }
+
+        // A concurrent join, or a setE2EEManager the app made while the
+        // resolver was still running, may have attached one already. That one
+        // holds the keys the session will use, so this is the surplus manager
+        // and it is the one that has to give its handle back.
+        if (_e2eeManager != null) {
+          _logger.d(() => '[resolveE2EE] a manager was attached meanwhile');
+          await manager.dispose().catchError((Object _) {});
```
renefloor: **Suggestion — empty catch on a key-store disposal.**

```dart
await manager.dispose().catchError((Object _) {});
```

This and the sibling at `call.dart:1181` (the surplus-manager and rollback paths) discard the error entirely, so a failed dispose leaks a native key store with zero record of it. `clearE2EEManager` gets this right at `call.dart:1233`:

```dart
await manager.dispose().catchError((Object e) {
  _logger.w(() => '[clearE2EEManager] dispose failed: $e');
});
```

Worth matching that here — swallowing silently is never quite right around a key store, and these two are the paths where a manager is being thrown away precisely because something already went wrong.

packages/stream_video/lib/src/call/session/call_session.dart
```diff
@@ -789,19 +798,27 @@ class CallSession extends Disposable {
 
       if (event is SfuJoinResponseEvent) {
         stateManager.sfuJoinResponse(event);
+        // The participant list just landed, so tracks that arrived before it
+        // can finally resolve a user id and attach their decryptor.
+        await rtcManager?.flushPendingDecryptors();
```
renefloor, line 803: **The decryptor wiring is still untested — the new tests cover the unit, not the integration.**

I re-ran the mutation experiment against `f1bc0cb9`. Commenting out all four wiring call sites:

- `unawaited(attachDecryptor(remoteTrack))` in `_onRemoteTrack` (`rtc_manager.dart:317`)
- all three `await rtcManager?.flushPendingDecryptors()` here (`:803`, `:806`, `:811`)

…leaves **all 536 tests passing.** The whole feature can still be deleted without a failure.

The reason is that `rtc_manager_decryptor_test.dart` invokes the methods directly — `manager.attachDecryptor(...)`, `manager.flushPendingDecryptors()`, `manager.discardPendingDecryptors(...)` at lines 116, 130, 133, 145, 148-149, 161, 165, 167, 194-195. Nothing drives them through `subscriber.onTrack` or through this event dispatch. So the queue/flush *logic* is now well covered (real progress — that part was previously untested too), but the claim "a late-arriving participant gets its decryptor" is still unproven end to end.

What's missing is one test per edge:
- a track delivered through `_onRemoteTrack` results in `attachDecryptor` being called at all;
- an `SfuJoinResponseEvent` / `SfuParticipantJoinedEvent` / `SfuTrackPublishedEvent` reaching `CallSession` triggers the flush;
- `SfuParticipantLeftEvent` triggers `discardPendingDecryptors`.

Worth doing precisely because the failure mode is silent: a permanently black participant with nothing in state or on the event stream (see my earlier comment on `attachDecryptor`). The wiring is the part most likely to be lost in a future refactor of this `if`-chain, and it's the part nothing currently protects.

Unrelated but adjacent, while you're in here: `SfuParticipantUpdatedEvent` (`:807`) still doesn't flush. Adding `SfuTrackPublishedEvent` covers the realistic case, so this is minor — but a participant update is what can first make a `trackIdPrefix` resolvable, and that path still waits for an unrelated event.

packages/stream_video/lib/src/call/call.dart
```diff
@@ -2639,17 +2922,37 @@ class Call {
 
   Future<void> _applyConnectOptions() async {
     _logger.d(() => '[applyConnectOptions] connectOptions: $_connectOptions');
-    await _applyCameraOption(
-      _connectOptions.camera,
-      _connectOptions.cameraFacingMode,
-      _connectOptions.targetResolution,
-      _connectOptions.videoInputDevice?.id,
+
+    void report(String option, Result<None> result) {
```
renefloor, line 2926: **Better, but still log-only — the app can't act on it.**

Threading `Result<None>` back out of `_applyCameraOption` / `_applyMicrophoneOption` / `_applyScreenShareOption` is the right plumbing, and `report()` means a refused publish is no longer *invisible*. That's a real improvement over discarding the `Result`.

But it still terminates in `_logger.e`. Nothing reaches `CallState`, no `Result` reaches the caller, no event fires. So the user-visible outcome is unchanged from what I described: on a transform-bridge hiccup during join, the user lands in a connected call — badged encrypted, per the still-open `isE2eeEnabled` issue — with camera and mic silently unpublished and mute state showing disabled. Indistinguishable from "I forgot to unmute", and unactionable by the app.

For the cross-SDK question that prompted this: Flutter's answer to "should we publish cleartext?" is now firmly *no*, verified. The answer to "or not publish at all and **inform the app** about it?" is still "not publish, and log about it."

Suggestion: surface it. A `CallState` field for "a track was refused because encryption could not be set up", or an event on the call, would let an app show something honest instead of a mystery muted state. The deliberate decision to fail closed is worth telling the user about — right now only the log knows.

packages/stream_video/lib/src/webrtc/e2ee/call_encryption_key.dart
```diff
@@ -0,0 +1,113 @@
...
+///
+/// Answer only for the calls you mean to encrypt, from whatever your app knows
+/// about them. The request carries the call's id and nothing else on purpose:
+/// the SDK does not tell you the call's encryption mode, because it does not
+/// reliably know it before the join and the other Stream SDKs do not offer one
+/// either. Returning a key for a call whose peers are not encrypting gets the
+/// join rejected by the server.
+///
+/// Return `null` for a call that needs no key from you. That is fine for a call
+/// whose encryption is `available`, which then joins unencrypted, and an error
```
renefloor, line 105: **Doc nit introduced by the mode removal — this sentence now describes behaviour that was deleted.**

```
/// Return `null` for a call that needs no key from you. That is fine for a call
/// whose encryption is `available`, which then joins unencrypted, and an error
/// for one that is `auto-on`, which cannot be joined without a key at all.
```

"an error … for one that is `auto-on`" described the old `required` branch in `_resolveE2EEManager`, which returned `Result.error('This call requires end-to-end encryption, but no key was available…')` before touching the network. That branch is gone in this commit — a `null` return now always succeeds locally and the join proceeds with `e2ee: false`, and the *server* rejects it.

The distinction matters to a resolver author: the failure now arrives later, as a coordinator join rejection with a server-worded message, rather than as an SDK error naming the fix. The paragraph you added just above already says this correctly ("Returning a key for a call whose peers are not encrypting gets the join rejected by the server") — these two are now inconsistent with each other.

Suggestion: "Return `null` for a call that needs no key from you. A call whose encryption is `available` then joins unencrypted; one that is `auto-on` cannot be joined without a key, and the server rejects the join."

packages/stream_video/lib/src/call/call.dart
```diff
@@ -1216,6 +1477,22 @@ class Call {
         SfuReconnectionStrategy.disconnect;
   }
 
+  /// Whether the coordinator refused the join in a way that will not change.
+  bool _isUnrecoverableCoordinatorError(VideoError? error) {
```
renefloor, line 1481: **Suggestion — this duplicates `RpcRetryManager._isRetryable`, inverted.**

The status-code policy here is the same one in `retry_manager.dart:87-93`, negated:

```dart
// call.dart
final status = cause.code;
if (status < 400 || status >= 500) return false;
return status != 401 && status != 408 && status != 429;

// retry_manager.dart
if (statusCode >= 400 && statusCode < 500) {
  return statusCode == 401 || statusCode == 408 || statusCode == 429;
}
```

Both also consult `error.apiError?.unrecoverable` first, with the same precedence. Two copies of a retry policy drift — and they drift silently, because "we retried something unretryable" and "we gave up on something retryable" are both quiet failures that only show up as odd reconnect behaviour in the field.

Suggestion: put the predicate in one place (e.g. alongside `apiError` in `video_error.dart`, as `bool get isUnrecoverable`) and have both call it. `VideoErrorApiDetails` is already the natural home, and both sites already have a `VideoError`.

Minor, same helper: `error!` at the `leave(reason: DisconnectReason.failure(error!))` call site is safe today only because `_isUnrecoverableCoordinatorError` returns `false` for `null`. That coupling is invisible at the call site — worth having the helper narrow the type, or asserting locally, so a future edit to the guard can't turn it into a crash on the failure path.

renefloor (commented):
### Re-review at `f1bc0cb9`

Re-ran everything against the two new commits. **536 tests pass** (up from 518) and `flutter analyze` is clean. Good progress — three of the five Criticals are properly fixed, not papered over.

#### Status of my earlier findings

| Finding | Status |
|---|---|
| `setE2EEManager` accepted mid-join → cleartext | ✅ **Fixed** |
| Encryption mode read before the coordinator knows it | ✅ **Fixed** (by removal) |
| `clearE2EEManager` disposes a shared, in-use manager | ✅ **Fixed** |
| Empty catches on `manager.dispose()` | ✅ **Fixed** |
| Failed publish left a track registered | ✅ **Fixed** (`_discardFailedPublish`) |
| dogfooding: fake random key on ringing joins | ✅ **Fixed** (now declines honestly) |
| `_stopTransceiver` swallowed its failure | 🟡 **Improved** — returns `bool`, caller escalates to `_logger.e`. Honest now, still log-only. |
| Failed decryptor attach invisible to app + event stream | 🟡 **Partial** — bookkeeping much better; the surfacing gap is untouched |
| Publish refusal not surfaced on the join path | 🟡 **Partial** — see inline |
| `isE2eeEnabled` unreconciled and never reset | ❌ **Open** — untouched |
| `join()` on a left `Call` leaks a manager + poisons the registry | ❌ **Open** — `_resolveE2EEManager` still runs at `call.dart:1307`, before the `_callLifecycleCompleter` check at `call.dart:1542` |
| Claim guard depends on GC timing | ❌ **Open** — still reads `claim.call.target` |

#### Verifying the three fixes

- **Mid-join attach:** the new status guard is genuinely sufficient. `lifecycleCallConnecting` fires at `call.dart:1547`, before the session is built (`call.dart:1649`) and before `joinCall(e2ee: ...)` (`call.dart:1847`). So any attach late enough to be missed is now rejected, and any attach early enough to be accepted is picked up by *both* the session and the join request. The window is closed, not narrowed.
- **Encryption mode:** dropping `encryptionMode` from the request is the right call, and the new doc is honest about why. It does trade a clear client-side error for a server rejection on `auto-on` without a key — a defensible call, since the server is the authority, but see the doc nit inline.
- **Shared manager:** removing the `dispose` flag *and* tightening the claimant check are coherent together; the two tests at `call_e2ee_test.dart:362` and `:383` now pin both directions.

#### The new decryptor tests close half the gap

`rtc_manager_decryptor_test.dart` is good work and covers the queue/flush logic properly. But I re-ran the mutation experiment and **it still passes with the production wiring deleted** — see the inline comment on `call_session.dart`.

`_stopTransceiver` returning `bool` deserves a note: with the encryptor race closed by the `_pendingTransceivers` gate (recorded in my previous review), a failed `stop()` is now the *only* remaining path to a live unencrypted sender. It's still log-only. Since it's now detectable, escalating it — failing the negotiation or triggering a reconnect rather than logging — would close the fail-closed chain completely.

Still open from the original pass, unchanged: `props` excludes `bytes` (`call_encryption_key.dart`), the platform check ordered after the resolver (`call.dart`), `StreamEncryptionMode.fromString`'s dead `default`, `call.dart:1037`'s `join(create: true)` snippet, and `random_words.dart:113` still using `Random()` rather than `Random.secure()` for the passphrase.

packages/stream_video/lib/src/call/call.dart
renefloor: Agreed that a left call cannot be re-joined — that's exactly the premise this depends on, so let me be concrete rather than argue, because I don't think the no-reuse rule closes it.

The rejection happens **downstream of the resolver**, and the cleanup path short-circuits. At `f1bc0cb9`:

1. `join()`'s preamble has no `_callLifecycleCompleter` guard. Its only status checks are `CallStatusConnected` (returns early), same-cid-in-progress, and Connecting/Joining (waits). A left call is `CallStatusDisconnected`, which passes all three.
2. `_resolveE2EEManager()` runs at `call.dart:1307` — so the resolver is invoked, `EncryptionManager.create` + `setSharedKey` run, and `setE2EEManager` registers the claim at `call.dart:1093`.
3. *Then* `_doJoin` rejects at `call.dart:1542` with `'call was left'`. Correct, per the no-reuse rule.
4. `join()`'s failure branch calls `leave()` → `_disconnect` hits `if (state.value.status.isDisconnected) return false`, which sits **before** the `try { … } finally { await _clear('disconnect'); }`. So `_clear` — and therefore `clearE2EEManager()` — never runs.

Net: the manager is never disposed, `_e2eeEventsSubscription` is never cancelled, and the claim stays in the static map pointing at a **live** `Call`, so the `WeakReference` never clears and the `removeWhere` sweep can't prune it. Every later `makeCall(cid).setE2EEManager(...)` for that cid then throws for the rest of the process.

I wrote a throwaway test against your branch to check I wasn't reasoning from a stale read:

```dart
final call = createTestCallWithState(
  initialState: CallState(
    preferences: DefaultCallPreferences(
      encryptionKeyResolver: (request) async {
        asked.add(request);
        return CallEncryptionKey.shared(bytes: Uint8List(16));
      },
    ),
    currentUserId: SampleCallData.defaultUserInfo.id,
    callCid: SampleCallData.defaultCid,
  ),
  coordinatorClient: setupMockCoordinatorClient(),
);

await call.leave();
expect(asked, isEmpty);        // passes — nothing resolved yet

final result = await call.join();
```

Output:

```
RESOLVER INVOCATIONS AFTER LEAVE: 1
JOIN RESULT: Result.Failure{error: VideoError{message: Could not set up end-to-end encryption:
  MissingPluginException(No implementation found for method initialize on channel FlutterWebRTC.Method)}}
```

So the resolver **is** invoked on a call that cannot be joined. Under `flutter test` it then fails at `EncryptionManager.create` because there's no platform channel — which is why `e2eeManager` reads `null` here and why the suite doesn't catch this. On a device that call succeeds, and steps 2-4 above play out in full.

If the intent is that this is unreachable in practice because apps don't re-join a left `Call`, that's fair as a likelihood judgement — but it's reachable by accident (a retry loop, a double-tap on a join button after a failure) and the consequence outlives the mistake, since the poisoned claim persists process-wide. A one-line `if (_callLifecycleCompleter.isCompleted) return const Result.success(none);` at the top of `_resolveE2EEManager` would close it without touching the API or the no-reuse rule.

Happy to be told this is a won't-fix — just wanted the trace on record rather than leaving it looking like I'd missed the reuse restriction.

packages/stream_video/lib/src/call/state/mixins/state_sfu_mixin.dart
renefloor: Fair on the naming — `isSfuE2eeNegotiated` and the derived getter were suggestions, and staying aligned with JS is a better reason to keep `isE2eeEnabled` than any reason I have to change it. Withdrawing both.

Two things in that comment are independent of naming and API shape, though, so I don't want them to disappear with the rename:

**The flag is never reset on teardown.** `lifecycleCallEnded` and `lifecycleCallDisconnected` both clear `sessionId` and `callParticipants` but not `isE2eeEnabled`. After leaving an encrypted call, `CallState` reads `isE2eeEnabled: true` while `e2eeManager` is already `null` — the two disagree, and a UI bound to the flag keeps its shield through the whole disconnected window. That's a one-line `copyWith(isE2eeEnabled: false)` in each, no rename and no new API, and I'd expect JS to clear its equivalent on leave too — worth checking rather than assuming it's aligned.

**Nothing carries decryption health.** Keeping the name means `isE2eeEnabled` stays "the SFU negotiated E2EE", which is fine — but then the question "is my media actually being encrypted/decrypted right now" has no answer anywhere in the SDK. `_onE2eeEvent` (`call.dart:1239`) already receives and severity-sorts `missingKey`, `decryptionFailed`, `encryptionFailed` and `unencryptedFrame`, and routes all of it to the logger. That's the padlock-lies case: manager attached, wrong key bytes at the right index, `isE2eeEnabled: true`, and zero decryptable frames.

If JS surfaces those events to the app in some form, matching that would settle it. If JS is also log-only, then this is a gap in both rather than a Flutter one — still worth raising with Oliver and Pratim rather than closing silently, since it's the failure users actually hit.

renefloor:
Thanks for going through these — responses to your three points.

**dogfooding security: withdrawn.** If the passphrase scheme is a cross-platform decision already made on pronto web, interop wins and there's nothing to fix — the derivation has to match what the other clients do. That covers the passphrase-in-URL note too. The only piece I'd still mention is `random_words.dart:113` using `Random()` rather than `Random.secure()`, since that one is local to Flutter and doesn't affect interop — but it's a demo passphrase generator that already documents itself as ~20 bits, so take it or leave it.

**Naming and API alignment with JS: agreed, withdrawn.** Replied on the `isE2eeEnabled` thread — the rename and the derived getter are dropped. Two points in that comment are independent of naming (the flag never being reset on disconnect, and decryption health having no representation anywhere), so I've kept just those, and flagged that both are worth *checking* against JS rather than assumed aligned.

**Manager leak on joining an already-left call: replied in thread with a trace and a test.** You're right that a left call can't be re-joined, and that's the premise — the issue is that the rejection lands *after* `_resolveE2EEManager` has already created and claimed the manager, and the `leave()` that follows short-circuits before `_clear` runs. I ran a throwaway test against `f1bc0cb9` to confirm the resolver really is invoked on a left call rather than reasoning from a stale read. Details in the thread; entirely reasonable to call it won't-fix on likelihood, I just wanted the trace on record.

For what it's worth, the three Criticals fixed in `a9d56e07`/`f1bc0cb9` are fixed properly — I checked that the mid-join guard is actually *sufficient* rather than just narrower, and the shared-manager fix is coherent across the API change and the tests.

packages/stream_video/lib/src/call/call.dart
```diff
@@ -1000,6 +1017,233 @@ class Call {
...
+      _logger.e(() => '[resolveE2EE] resolver threw: $e; $stk');
+      return Result.error('The encryption key resolver failed: $e');
+    }
+
+    if (key == null) {
+      _logger.d(() => '[resolveE2EE] no key, joining unencrypted');
+      return const Result.success(none);
+    }
+
+    if (!EncryptionManager.isSupported) {
```
renefloor, line 1130: **Reorder this ahead of the resolver — as written it breaks every join on web, Windows and Linux.**

`CallPreferences.encryptionKeyResolver`'s doc recommends setting this on `StreamVideoOptions.defaultCallPreferences` to cover every call. An app that follows that advice and returns a key unconditionally now fails **every** join on any platform where `EncryptionManager.isSupported` is `false` — including calls whose mode is `disabled` or `available`, which would otherwise join fine unencrypted.

The resolver has no way to avoid it: the request carries only `callCid`, so a resolver can't tell an unsupported platform from a supported one without calling `EncryptionManager.isSupported` itself, which the docs don't mention.

Removing the `required` distinction made this sharper rather than milder — a resolved key on an unsupported platform is now unconditionally fatal.

Suggestion: check support *before* invoking the resolver and return `Result.success(none)` (join unencrypted), or keep the call and treat unsupported as "no key". Either way an `auto-on` call still fails, but at the server rather than locally, which is consistent with how the mode is handled everywhere else now.

packages/stream_video/lib/src/call_state.dart
```diff
@@ -165,6 +167,9 @@ class CallState extends Equatable {
   /// Web only; always `false` on every other platform.
   final bool isWebAudioPlaybackBlocked;
 
+  /// Whether the SFU reports this call as end-to-end encrypted.
+  final bool isE2eeEnabled;
```
renefloor, line 171: **This is never reset, so it survives the call it describes.**

To be clear on scope: I've withdrawn the rename and the derived getter — staying aligned with JS is a better reason to keep `isE2eeEnabled` than any reason I had to change it. This is the correctness half, and it's independent of naming or API shape.

The field is written only by `sfuE2eeEnabledUpdated`, from the join response and fast-reconnect. `lifecycleCallEnded` (`state_lifecycle_mixin.dart:58`) and `lifecycleCallDisconnected` (`:181`) both clear `sessionId` and `callParticipants` but not this. So after leaving an encrypted call, `CallState` reports `isE2eeEnabled: true` while `e2eeManager` is already `null` — the two contradict each other and neither is authoritative.

Invisible in dogfooding, since the badge in `call_duration_title.dart` only renders in-call. But it's a security indicator, and an SDK consumer binding a shield to it shows it through the entire disconnected window.

Suggestion: `copyWith(isE2eeEnabled: false)` in both. Worth confirming JS clears its equivalent on leave rather than assuming the stickiness is the aligned behaviour.

renefloor (commented):
### Two small fixes worth landing before merge

Narrowing to the two I'd actually gate on — both are self-contained and neither touches the public API shape. Everything else I've raised is either already answered, a judgement call that's yours, or reasonable follow-up work.

renefloor (approved):

PR #1320, opened 2026-09-01
feat(ui, samples): lobby view

renefloor:
### 🎯 Goal

Update lobby view and unify sdk/dogfooding.

### 🛠 Implementation details

Added 3 main options for call controls, with small, medium and large.

The simple buttons are mostly intended for android/ios. The full buttons are meant for web. Regular is something that can be used on both. Simple buttons only have on/off for camera and mic. Full have the same on/off, but also the new input selections. The regular one is a bit in between using the new split button, so getting an option to switch inputs, without the large UI. This PR also added a split between CallControlButton and CallFeatureButton. The controls are for things like mic/camera that should be on/off. The Features are for things like screen sharing and more menu, things that are off by default but are sometimes active.

The `LobbyActions` have controls and settings. The controls can be shown on the camera feed on wide screens, but are always below the feed if there are more than 3 controls to prevent overlap with the participant label.
The helper constructors like `LobbyActions.simple` have the option to add more controls or settings.

Screenshots from macos style snapshot tests

| simple | regular | full | with extras |
| --- | --- | --- | --- |
| stream_lobby_view_simple_light | stream_lobby_view_regular_light | stream_lobby_view_full_light | stream_lobby_view_extras_light |

The input selection uses an AnchorMenu on desktop/web, but a bottom sheet on android/ios.
It works the same in the lobby and the call screen

| macos | iPad | iPhone |
| --- | --- | --- | 
| image | image | image |
| image | image | image | 
| image | image | image |

#### General

### ☑️Reviewer Checklist
- [ ] Sample runs & works


dogfooding/ios/Flutter/AppFrameworkInfo.plist
```diff
@@ -20,7 +20,5 @@
   <string>????</string>
   <key>CFBundleVersion</key>
   <string>1.0</string>
-  <key>MinimumOSVersion</key>
-  <string>15.0</string>
```
renefloor, line 24: revert this

dogfooding/lib/screens/call_screen.dart
```diff
@@ -338,45 +393,95 @@ class _CallScreenState extends State<CallScreen> {
...
+                                builder: (context, enabled) =>
+                                    StreamMicrophoneSplitButton(
+                                      devices: _devices,
+                                      enabled: enabled,
+                                      unavailable: _noDeviceFor(
+                                        _devices.audioInputs,
+                                      ),
+                                      // The bar sits along the bottom, so its
+                                      // menus come up rather than down.
+                                      menuDirection: StreamMenuDirection.up,
```
renefloor: I think we can simply do 

```suggestion
                                      menuDirection: .up,
```

with the current min dart language

dogfooding/lib/screens/call_screen.dart
```diff
@@ -338,45 +393,95 @@ class _CallScreenState extends State<CallScreen> {
...
+                                            ),
+                                    ),
+                              ),
+                            ],
+                          ),
                         ),
                         const Spacer(),
-                        PartialCallStateBuilder(
+                        // onTap, so the button opens this app's own
+                        // participants screen rather than the SDK's list.
```
renefloor: No need for these comments

dogfooding/lib/screens/call_screen.dart
```diff
@@ -338,45 +393,95 @@ class _CallScreenState extends State<CallScreen> {
...
+                        PartialCallStateBuilder<List<CallParticipantState>>(
```
renefloor: Why do we need this typedef now? `selector: (state) => state.callParticipants,` already gives the type.

packages/stream_video_flutter/lib/src/call_controls/participants_control.dart
```diff
@@ -0,0 +1,101 @@
...
+/// {@tool snippet}
+///
+/// ```dart
+/// StreamParticipantsControl(
+///   participants: participants,
+///   onTap: () => openParticipantsPanel(),
+/// )
+/// ```
+/// {@end-tool}
+class StreamParticipantsControl extends StatelessWidget {
```
renefloor: We renamed all Controls to Buttons, but this is still a Control

packages/stream_video_flutter/lib/src/call_participants/screen_share_call_participants_content.dart
```diff
@@ -95,7 +95,7 @@ class ScreenShareCallParticipantsContent extends StatelessWidget {
                   mainAxisSize: MainAxisSize.min,
                   children: [
                     Icon(
-                      Icons.screen_share_rounded,
+                      context.streamIcons.presentDesktopFill,
```
renefloor: The icon should depend on the platform desktop/mobile.

packages/stream_video_flutter/lib/src/call_screen/lobby_actions/lobby_participants_control.dart
```diff
@@ -0,0 +1,37 @@
...
+import '../../../stream_video_flutter.dart';
+
+/// The lobby's participants control: [StreamParticipantsControl] over the
+/// people already in the call.
+///
+/// Exists so a preset can list it without wiring anything up. Use
+/// [StreamParticipantsControl] directly anywhere there is no
+/// [StreamLobbyScope] — in a call, where an app usually wants `onTap` to open
+/// a panel of its own rather than the built-in list.
+class StreamLobbyParticipantsControl extends StatelessWidget {
```
renefloor: Same, should not be a control

packages/stream_video_flutter/lib/src/call_screen/lobby_controller.dart
```diff
@@ -0,0 +1,663 @@
...
+/// Everything the lobby knows before the call is joined.
+///
+/// The call is read, not created: one that does not exist yet leaves
+/// [fetchError] set and [participants] empty, and creating it is the host's to
+/// do — before the lobby, or from [StreamLobbyView.onJoinCallPressed].
+class StreamLobbyController extends ChangeNotifier {
+  /// Creates a new instance of [StreamLobbyController].
+  StreamLobbyController({
+    required this.call,
+    StreamVideo? streamVideo,
```
Brazol: I'm not sure about passing StreamVideo instance as a parameter to LobbyView and/or LobbyViewController. Do we do something like this in any other place in the SDK?

packages/stream_video_flutter/lib/src/call_screen/lobby_controller.dart
```diff
@@ -0,0 +1,663 @@
...
+    }
+  }
+
+  void _fetchCall() {
+    // Reads the call, and does not create one that is not there: what a call
+    // is created with — its encryption mode above all, which cannot be changed
+    // afterwards — is the host's to decide, and a waiting room being shown is
+    // not that decision. A call that does not exist yet fetches as a failure,
+    // which leaves the lobby with a preview, no participants and [fetchError].
+    final currentUserId = _video.currentUser.id;
```
Brazol: We could get current user from call state instead of video instance: `call.state.value.currentUserId`

packages/stream_video_flutter/lib/src/call_screen/lobby_controller.dart
```diff
@@ -0,0 +1,663 @@
...
+    if (_disposed) return;
+
+    _logger.e(() => '[fetchCall] threw: $error\n$stackTrace');
+    _fetchError = error;
+    _applyCallDefaults(call.state.value.settings);
+    _notify();
+  }
+
+  void _listenEvents() {
+    _eventSubscription?.cancel();
```
Brazol, line 550: For events listening maybe we could use the `_init()` method like it is done in ringing flow to start listening easier (before `join()`) when lobby is shown? We could remove the StreamVideo instance dependency in lobby then

packages/stream_video_flutter/lib/src/call_screen/lobby_controller.dart
```diff
@@ -0,0 +1,663 @@
...
+    _eventSubscription = _video.events.listen(
```
Brazol: If StreamVideo instance stays then we should filter the events by callId.

packages/stream_video_flutter/lib/src/call_screen/lobby_controller.dart
renefloor: No, nowhere else — this was the only widget in the package holding a `StreamVideo`. It's gone from both `StreamLobbyView` and `StreamLobbyController` now: the local user comes off the call, and so do the events. https://github.com/GetStream/stream-video-flutter/commit/83d7d77bc42cb054d31ad76c696fdf1665ce15a1

packages/stream_video_flutter/lib/src/call_screen/lobby_controller.dart
renefloor: Done for the two id comparisons. The third use needed more: `currentUser` feeds the preview tile's name, image and role, and `CallState` carries only the id — so I added `Call.currentUser` in the LLC, delegating to the client the call already holds. https://github.com/GetStream/stream-video-flutter/commit/83d7d77bc42cb054d31ad76c696fdf1665ce15a1

packages/stream_video_flutter/lib/src/call_screen/lobby_controller.dart
renefloor: Right idea, and it turned out to need nothing new: `Call.get` defaults to `watch: true`, and `_performGetOperation` already calls `_observeEvents()`, so the lobby's own fetch is what makes `call.callEvents` live before the join. The lobby listens there now and no longer needs the client.

I stayed away from `_init()` itself: it also wires `_observeReconnectEvents()`, and `_reconnect` only bails on `CallStatusDisconnected` — so a lobby-initialized call would try to reconnect a call nobody joined when the network drops.

Added an LLC test for the pre-join behaviour, since the lobby now stands on it. https://github.com/GetStream/stream-video-flutter/commit/83d7d77bc42cb054d31ad76c696fdf1665ce15a1

packages/stream_video_flutter/lib/src/call_screen/lobby_controller.dart
renefloor: Good catch — it was a real bug, and an old one: the released lobby has the same unfiltered listen. Filtered by cid now, so a call ringing in the background no longer moves people in and out of this lobby's list. Covered by a test (verified it fails without the filter). https://github.com/GetStream/stream-video-flutter/commit/83d7d77bc42cb054d31ad76c696fdf1665ce15a1

Brazol (approved):

PR #1323, opened 2026-09-02
feat(llc): Android telecom integration

Brazol:
## Summary by CodeRabbit

- **New Features**
  - Added optional Android Telecom integration for incoming and outgoing calls, including call controls, audio routing, and Android 17 ringing support.
  - Added Telecom configuration options.
  - Added iOS call-history reporting with selectable end reasons.
  - Added events identifying when iOS cannot display an incoming call, including failure details.

- **Bug Fixes**
  - Calls ended through watches, headsets, or vehicle systems now properly leave the Stream call.
  - Improved cold-start incoming-call handling and CallKit transaction failure handling.

- **Documentation**
  - Expanded Android Telecom and iOS call-history guidance.


packages/stream_video_push_notification/android/build.gradle
```diff
@@ -45,6 +45,8 @@ android {
         implementation 'com.fasterxml.jackson.core:jackson-core:2.20.0'
         implementation 'com.fasterxml.jackson.core:jackson-databind:2.20.0'
         implementation "io.coil-kt:coil:1.4.0"
+        implementation 'androidx.core:core-telecom:1.0.1'
```
renefloor, line 48: **`core-telecom` silently adds `BLUETOOTH_CONNECT` to every consuming app, and integrators need to request it or ringing may break.**

I built `packages/stream_video_push_notification/example` and read the manifest-merger report:

```
uses-permission#android.permission.BLUETOOTH_CONNECT
ADDED from [androidx.core:core-telecom:1.0.1] .../jetified-core-telecom-1.0.1/AndroidManifest.xml:22
```

The AAR also merges in two exported components (`androidx.core.telecom.internal.JetpackConnectionService`, `MuteStateReceiver`), and `MODIFY_AUDIO_SETTINGS` — that last one already merged from `audioswitch`, so it's not new. `BLUETOOTH_CONNECT` is.

The part that matters for integrators: `BLUETOOTH_CONNECT` is a **dangerous runtime permission**. Merging it into the manifest does not grant it. Apps that don't already request it at runtime will have it declared and denied, and Jetpack Telecom's endpoint enumeration needs it — so a `SecurityException` during `addCall` would fail registration, and on Android 17 a failed registration means the ringtone is silently muted (per this PR's own rationale for defaulting Telecom on there). That chain is worth verifying on-device, but it means **integrators must add a runtime request for `BLUETOOTH_CONNECT` or ringing can break on 17** — which needs to be called out in the README and CHANGELOG, not discovered.

For reference, both other SDKs own this permission explicitly rather than inheriting it invisibly:
- **stream-video-android** declares it in `stream-video-android-core/src/main/AndroidManifest.xml:29` and ships Compose helpers to request it, gated on API 31+ (`CallPermissions.kt:110-116`, `SinglePermission.kt:114`).
- **react-native** hits the identical silent merge (`react-native-callingx` depends on the same `core-telecom:1.0.1`), but on top of that adds `BLUETOOTH`/`BLUETOOTH_CONNECT`/`BLUETOOTH_ADMIN` via its Expo config plugin (`withAndroidPermissions.ts:7-12`), documents them in the plugin README, and gates usage at runtime — `BluetoothManager.kt:210` checks the grant and `start()` bails with `Log.w(TAG, "App lacks BLUETOOTH permission")` so the audio stack degrades cleanly.

Minimum: README "Requirements and behaviour" list + CHANGELOG. Better: also degrade gracefully and log when the grant is missing.

---

Separately on this line: **`core-telecom:1.0.1` declares `minSdkVersion="23"`** but the plugin still declares `minSdk = 21` on line 36. Flutter's current default is 24 so a default project merges fine, but any app explicitly on 21/22 fails manifest merge. Worth raising the plugin's `minSdk` to 23 so the declaration is honest, and noting it under `### 🔄 Changed` next to the Java 17 bump.

packages/stream_video_push_notification/android/src/main/kotlin/io/getstream/video/flutter/stream_video_push_notification/StreamVideoPushNotificationPlugin.kt
```diff
@@ -293,6 +309,44 @@ class StreamVideoPushNotificationPlugin: FlutterPlugin, MethodCallHandler, Activ
...
+                        StreamTelecomManager.disconnectAll(it, DisconnectCause.LOCAL)
+                    }
+
+                    result.success(true)
+                }
+
+                // Ends the ring, not the call: the user accepted and is in the call now, so the
+                // notification and ringtone go but the Telecom registration stays. Kept apart from
+                // "endAllCalls" because an ENDED broadcast alone cannot say which of the two it is.
+                "dismissRingingCall" -> {
```
renefloor, line 326: **This ignores the `cid` Dart sends.** The method channel sends `{'id': cid}` (`stream_video_push_notification_method_channel.dart:137`), but the handler never reads `call.arguments` — it iterates *every* entry from `getDataActiveCalls`, answers each in Telecom, broadcasts ENDED with `keepSystemCall = true` for each, then `removeAllCalls`.

With two stacked ringing calls, accepting one answers the other in Telecom and dismisses its ring. Because `keepSystemCall = true` suppresses the disconnect and `removeAllCalls` wipes it from `ACTIVE_CALLS`, the result is a phantom active call in the system call state (live on watch/car UI) that no Dart code can subsequently address — the `endCallByCid` → `endAllCalls` sweep won't find it either.

`"endCall"` at :255 does this correctly with `firstOrNull { it.id == data.id }`. Two smaller notes: every other method on this channel puts a **uuid** under `'id'`, not a cid, so the contract is inconsistent here; and the dropped argument isn't logged, so nothing signals it.

packages/stream_video_push_notification/android/src/main/kotlin/io/getstream/video/flutter/stream_video_push_notification/telecom/StreamTelecomManager.kt
```diff
@@ -0,0 +1,497 @@
...
+
+    private var appContext: Context? = null
+    private var config: TelecomConfig = TelecomConfig.DISABLED
+
+    /** Guards the one-off read of the persisted configuration. */
+    @Volatile
+    private var isConfigResolved: Boolean = false
+
+    @SuppressLint("NewApi")
+    private var repository: TelecomCallRepository? = null
```
renefloor, line 51: **Needs `@Volatile`** — this is unsafe double-checked locking. It's read outside the lock at :126 (`repository?.let { return it }`) and written inside `synchronized(this)` at :127-138, plus unsynchronized reads at :258, :271, :300, :345 and :354. Readers genuinely span threads: main-thread broadcast handlers vs `Dispatchers.Default` coroutines launched from `init()`/`register()`. Without the barrier a caller can observe a partially-constructed `TelecomCallRepository`.

`isConfigResolved` right below is correctly volatile, so this looks like an oversight rather than a decision. Same applies to `IncomingCallBroadcastReceiver.silenceEvents`, which is now read from background coroutines at :487.

renefloor (commented):
Review of the Telecom integration. Overall this is careful work — the error handling in `TelecomCallRepository` is genuinely good (re-throwing `CancellationException` before the general catch, checking `CallControlResult.Error` on every control call, disconnecting with `BUSY` on a failed answer rather than leaving the call half-answered). It also fixes a real bug the Android SDK has: `JetpackTelecomRepositoryProvider` there uses bitwise `and` on the register capabilities (`1 shl 1` and `1 shl 2`, so it registers with zero capabilities); this uses `or`. It supports concurrent calls where the Android SDK bails on the second, and activates outgoing calls on scope-ready instead of the Android SDK's 2s `delay()`.

Inline comments below. Three things that don't map to a diff line:

**CI does not compile any of this PR's Kotlin.** `melos run build:example:android` builds `packages/stream_video_flutter/example`, which does not depend on `stream_video_push_notification` at all (zero hits in its manifest-merger report). ~1500 lines of new Kotlin, no Kotlin unit tests, and no compile coverage in CI. Building `packages/stream_video_push_notification/example` instead — or in addition — would fix that cheaply.

**`init()` is unawaited with no error handling** — `lib/src/stream_video_push_notification.dart:254`, a bare `Future<void>` in a synchronous factory closure. This PR newly routes Android through it (`isIos` → `isMobile` in the method channel), and the Kotlin handler converts every exception into `result.error`. A failure means `StreamTelecomManager.init` never runs and the config is never persisted, so on Android <17 an explicit `enabled: true` is silently discarded — surfacing only as an unhandled zone error that never mentions push.

**Stale PhoneAccount on upgrade.** On `main`, `InAppCallManager.registerPhoneAccount()` ran on every plugin attach. This PR deletes `InAppCallManager` and `IncomingCallConnectionService` without ever calling `unregisterPhoneAccount`, so existing installs upgrade with a registered self-managed PhoneAccount pointing at a ComponentName that no longer exists. `PhoneAccountRegistrar` usually prunes those on package update, so it likely self-heals — worth confirming on-device that it doesn't collide with the new `registerAppWithTelecom` account. Also worth a `### 🔄 Changed` line, since `InAppCallManager` was a public class.

packages/stream_video_push_notification/android/src/main/kotlin/io/getstream/video/flutter/stream_video_push_notification/telecom/StreamTelecomManager.kt
```diff
@@ -0,0 +1,497 @@
...
+        else -> "available"
+    }
+
+    @SuppressLint("NewApi", "MissingPermission")
+    private fun register(context: Context, data: Data, incoming: Boolean) {
+        if (!isEnabled(context)) {
+            Log.d(TAG, "[register] skipped: ${describeAvailability(context)}")
+            return
+        }
+        val repository = requireRepository(context) ?: return
```
renefloor, line 175: **All of this runs on the main thread inside a `BroadcastReceiver`.** `IncomingCallBroadcastReceiver.onReceive` → `register()` → `isEnabled(context)` → `ensureInitialized` → `TelecomConfig.load` (SharedPreferences disk read) and `TelecomPermissions.supportsTelecom` → `defaultDialerPackage` (binder IPC) → `requireRepository` → `TelecomCallRepository(...).warmUp()` → `CallsManager.registerAppWithTelecom` (another binder round-trip).

This is exactly the cost the comment at :95-96 says the `init()` warm-up exists to avoid — but on a cold start from push, `init()` hasn't run yet, so the very first ring pays all of it synchronously in a receiver (10s limit, StrictMode disk + IPC violations), which is the worst possible moment.

Moving the `isEnabled` / `requireRepository` work inside the existing `scope.launch` below would fix it.

packages/stream_video_push_notification/android/src/main/kotlin/io/getstream/video/flutter/stream_video_push_notification/telecom/StreamTelecomManager.kt
```diff
@@ -0,0 +1,497 @@
...
+                IncomingCallNotificationService.repromoteRingingForeground()
+            } else {
+                // An outgoing call stays in the dialing state until it is explicitly activated.
+                // The Android SDK does this on a fixed delay after registering; doing it once the
+                // scope is actually up avoids racing that timer.
+                processAction(callId, TelecomCallAction.Activate)
+            }
+        }
+
+        override fun onRegistrationFailed(callId: String, error: Throwable?) {
```
renefloor, line 398: **On Android 17 this fallback means a silent incoming call, and nothing is told about it.**

The log line says "falling back to notification-only flow", but per this PR's own rationale (`TelecomConfig.defaultEnabled()`, the README table) the platform won't play a ringtone from a push-started service unless the call is in the Telecom stack. So on 17 "notification-only" isn't degraded ringing — it's no ringing. The only trace is a `Log.w`.

Same applies to the skip path at :152-160: `describeAvailability` produces genuinely useful text ("MANAGE_OWN_CALLS not granted", "no telephony feature or default dialer") and then throws it away at `Log.d`, which most production log pipelines strip.

This PR just added `ActionCallIncomingFailed` for the iOS equivalent — the Android side has the same condition with no event. Worth emitting one here (and in `promoteToForeground`'s catch) so integrators can detect it. react-native-callingx does this: `CALL_REGISTRATION_FAILED` broadcast → rejected JS promise with a `REGISTRATION_FAILED` code, plus a 10s registration timeout.

packages/stream_video_push_notification/android/src/main/kotlin/io/getstream/video/flutter/stream_video_push_notification/telecom/StreamTelecomManager.kt
```diff
@@ -0,0 +1,497 @@
...
+            // A call also becomes active on connect and on answer, both of which mean it can no
+            // longer be disconnected as a decline.
+            connectedCallIds.add(callId)
+            val wasHeld = heldCallIds.remove(callId)
+            if (selfRequestedHold.remove(callId, false)) return
+            if (!wasHeld) return
+            sendEvent(IncomingCallConstants.ACTION_CALL_TOGGLE_HOLD, callId, "isOnHold" to false)
+        }
+
+        override fun onIsCallInactive(callId: String) {
```
renefloor, line 464: **`ACTION_CALL_TOGGLE_HOLD` is emitted but nothing on the Dart side handles it.** `ActionCallToggleHold` appears in Dart only inside a `toString` at `stream_video_push_notification.dart:717`.

So when a cellular call arrives mid-Stream-call, Telecom calls `onSetInactive`, the OS takes the audio route, and the Stream call keeps publishing with the app unaware. The asymmetry looks accidental: `ActionCallToggleMute` two handlers up *is* wired to `call.setMicrophoneEnabled` at `stream_video_push_notification.dart:202`.

stream-video-android handles this by leaving the call outright (`CallState.kt:1871-1888`, `SdkCause.CALL_ON_HOLD`) — aggressive, but at least defined. react-native has the same gap, so this is a consistent cross-platform hole rather than a Flutter regression, but it needs either an owner or an explicit "integrator's responsibility" note in the README.

(Also: app→platform mute never reaches Telecom either — `setCallMutedByCid` early-returns off iOS at `stream_video_push_notification.dart:496`.)

packages/stream_video_push_notification/android/src/main/kotlin/io/getstream/video/flutter/stream_video_push_notification/telecom/StreamTelecomManager.kt
```diff
@@ -0,0 +1,497 @@
...
+            // Leaving the flag set would swallow the next system-originated hold on this call, and
+            // clearing one that belongs to a different request would do the same.
+            selfRequestedHold.remove(callId, requestedOnHold)
+        }
+
+        override fun onCallAudioEndpointsChanged(callId: String, snapshot: Map<String, Any?>) {
+            sendEvent(
+                IncomingCallConstants.ACTION_CALL_CUSTOM,
+                callId,
+                "type" to CUSTOM_EVENT_AUDIO_ENDPOINTS,
```
renefloor, line 486: **The audio endpoint snapshot is emitted as an untyped `ActionCallCustom` and nothing consumes it** — I grepped `stream_video/lib`, `stream_video_push_notification/lib` and `stream_video_flutter/lib`; there is no `ActionCallCustom` handler anywhere. There's also no action to *request* an endpoint change: `TelecomCallAction` has only `Answer`/`Disconnect`/`Hold`/`Activate`.

The consequence is two owners fighting over audio routing with no arbitration. `Call.setAudioOutputDevice` (`call.dart:3833`) still goes to `rtc.Helper.selectAudioOutput`, i.e. straight into flutter-webrtc's `AudioManager`/SCO stack, which Telecom overrides. This is precisely the bug stream-video-android wrote a workaround for — `MediaManager.kt:669-675`:

> *"the Telecom framework controls audio routing and has higher priority than AudioManager.setCommunicationDevice(). Route the switch through the Telecom API so the OS honours it on all devices including Samsung S21 Ultra (Android 14)."*

The two references solve it in opposite directions — Android routes selections *into* Telecom (`switchTelecomEndpointForDevice`, `MediaManager.kt:692-728`); react-native suppresses its own audio stack when Telecom owns the call (`setTelecomManagedMode`, `AudioDeviceManager.kt:102-168`), which is arguably the cleaner of the two. Either works; doing neither doesn't.

This is the one I'd push hardest on, and it bites hardest for **video** calls — which this PR enables Telecom for by default from API 37, a configuration neither reference ships (the Android SDK's demo enables Telecom only for `CallType.AudioCall`).

Minor, same area: `observeCalls` only fires `onCallRegistered` when `previous == null`, so the *initial* endpoint and mute snapshot never reaches Dart.

packages/stream_video_push_notification/lib/src/stream_video_push_configuration.dart
```diff
@@ -85,6 +86,9 @@ class AndroidPushConfiguration {
   /// Show full locked screen.
   final bool? showFullScreenOnLockScreen;
 
+  /// Android Telecom integration. Disabled unless explicitly provided.
```
renefloor: This contradicts the rest of the PR. `TelecomConfig.fromConfiguration` treats an absent `telecom` section as "no preference" and falls through to `defaultEnabled()`, which is **on** from API 37 — as the `enabled` docstring below, the CHANGELOG and the README table all say.

Suggest something like: `/// Android Telecom integration. Follows the platform default when omitted — see [TelecomPushConfiguration.enabled].`

packages/stream_video_push_notification/ios/stream_video_push_notification/Sources/stream_video_push_notification/StreamVideoCallkitManager.swift
```diff
@@ -252,15 +266,19 @@ public class StreamVideoCallkitManager: NSObject, CXProviderDelegate {
         guard let uuid = UUID(uuidString: data.uuid) else { return }
```
renefloor: **This `return` skips the PushKit `completion()`.** Pre-existing, but this PR reworked this exact closure and correctly added `completion()` to the new error branch, so it's worth closing the remaining escape while you're here. The `sharedProvider?.` optional chain is a second one — if it's nil the whole call *and* its closure are skipped, though `initCallkitProvider(data)` on the line above makes that unlikely in practice.

Missing the PushKit completion is not a soft failure: iOS terminates the app and eventually stops delivering VoIP pushes to it. A malformed `data.uuid` also produces no `ACTION_CALL_INCOMING_FAILED` and no log — the call just vanishes.

Suggest logging, sending `ACTION_CALL_INCOMING_FAILED` with a synthetic reason, and calling `completion()` on both escapes.

packages/stream_video_push_notification/ios/stream_video_push_notification/Sources/stream_video_push_notification/StreamCallKitCallController.swift
```diff
@@ -125,18 +127,41 @@ class StreamCallKitCallController: NSObject {
     private func requestCall(
```
renefloor, line 126: This refactor is a real improvement, and `clearCallsAfterFailedEnd` below is the right escape hatch for a refused end transaction.

One gap: **`connectedCall` (:74-88) still bypasses this helper** and issues its `CXAnswerCallAction` directly with `print`-and-drop error handling. A refused answer transaction leaves the CallKit screen ringing while the app believes it's connected — the same class of stuck UI `clearCallsAfterFailedEnd` exists to prevent. Routing it through `requestCall(_:action:completion:)` would cover it.

packages/stream_video_push_notification/test/telecom_configuration_test.dart
```diff
@@ -0,0 +1,99 @@
...
+
+      expect(
+        TelecomPushConfiguration.fromJson(config.toJson()),
+        isA<TelecomPushConfiguration>()
+            .having((c) => c.enabled, 'enabled', isTrue)
+            .having((c) => c.schema, 'schema', 'myapp'),
+      );
+    });
+
+    test('is carried by the android section of the push configuration', () {
```
renefloor: This asserts on a hand-built `toJson()`, which skips the part that can actually break. The production path is `_defaultPushConfiguration.merge(...)` → `configuration.toJson()` → `StreamVideoPushNotificationPlatform.instance.init(...)` → the method channel's `if (!CurrentPlatform.isMobile) return`.

That `isIos` → `isMobile` change is the load-bearing line of the whole Android feature — before it, `initData` never reached Android and `StreamTelecomManager.init` never ran. **Revert that one line and every test in this PR still passes.**

A fake `StreamVideoPushNotificationPlatform` that captures the map handed to `init`, driven through `StreamVideoPushNotificationManager.create` with a telecom-bearing config, would cover the merge, the serializer and the dispatch in one test. The pattern already exists in this directory at `ringing_event_broadcaster_test.dart:8-19`.

Two other gaps worth a case each:
- `IncomingCallFailureReason.fromName` is untested and matches by **enum member name** against Swift string literals (`StreamVideoCallkitManager.swift:421-434`), degrading silently to `unknown` on any rename or typo. That's the higher-risk of the two enums; the PR tests the lower-risk `CallEndedReason` instead.
- `endedBySystem` is untested end to end, despite changing call-teardown semantics for every Android user at `stream_video.dart:1221`.

Also, three cases here can't fail for the right reason: :23 tests `json_serializable`'s handling of a nullable bool, :56 asserts `null == null` (the observable case — base has telecom, override doesn't — isn't covered), and :90 restates an enum declared in this same PR.

renefloor (approved):

PR #1324, opened 2026-09-03
feat(ui): call controls

renefloor:
### 🎯 Goal

Making new call control bar

### 🛠 Implementation details

This adds a new responsive call controls bar with no controls by default.

### 🎨 UI Changes

https://github.com/user-attachments/assets/5d4ecc3b-ebb4-4ff2-8509-a606a67bea0f

#### General

### ☑️Reviewer Checklist
- [x] Sample runs & works
- [x] UI Changes correct (before & after images)
- [x] Bugs validated (bugfixes)
- [x] New feature tested and works
- [x] All code we touched has new or updated Documentation


packages/stream_video_flutter/lib/src/call_controls/controls/stream_layout_button.dart
```diff
@@ -3,9 +3,9 @@ import 'package:flutter/material.dart';
...
 /// A widget that allows to change the layout of the call.
-class ToggleLayoutOption extends StatefulWidget {
-  /// Creates a new instance of [ToggleLayoutOption].
-  const ToggleLayoutOption({
+class StreamLayoutButton extends StatefulWidget {
+  /// Creates a new instance of [StreamLayoutButton].
+  const StreamLayoutButton({
     super.key,
     required this.onLayoutModeChanged,
     this.initialLayout = ParticipantLayoutMode.grid,
```
renefloor: Why is this called `initialLayout` and not just `layout`. It's not that the button keeps track of layout changes.

dogfooding/lib/screens/call_screen.dart
```diff
@@ -373,119 +431,68 @@ class _CallScreenState extends State<CallScreen> {
                 );
               },
               callControlsWidgetBuilder: (BuildContext context, Call call) {
```
renefloor: Better to have this whole builder in a separate method

packages/stream_video_flutter/lib/src/call_controls/call_control_bar.dart
```diff
@@ -0,0 +1,315 @@
...
+///  * [CallControlBarTheme], for restyling the bar in a subtree.
+class CallControlBar extends StatelessWidget {
+  /// Creates a control bar.
+  const CallControlBar(
+    this.small, {
+    super.key,
+    this.medium,
+    this.large,
+    this.style,
+    this.primary = true,
```
renefloor: Better to order this differently, no key is in between small and medium. Put `super.key` last.

packages/stream_video_flutter/lib/src/call_screen/media_devices_controller.dart
```diff
@@ -137,6 +137,15 @@ class StreamMediaDevicesController extends ChangeNotifier {
...
   final bool supportsSystemDefault;
 
+  /// The id a platform gives the device it has picked for itself.
+  ///
+  /// Web reports one — Chrome lists it as "Default - <name>" — alongside the
+  /// real devices, and it is a device like any other: picking it is what
+  /// asking for the platform's choice looks like there. Platforms that report
+  /// no such entry leave [selectedAudioInput] and its siblings null until
+  /// something is picked.
+  static const platformDefaultDeviceId = 'default';
```
Brazol, line 181: I'm not sure how this "defaults" are used exactly but shouldn't this be platform specific? The 'default' value might work for web but on mobile devices the default is probably 'earpiece' or 'speaker'.

Brazol (approved):

packages/stream_video_flutter/lib/src/call_screen/media_devices_controller.dart
renefloor: Good catch — you are right, and it was web-only.

Only a browser reports the reserved `default` id (Chrome lists it as "Default - <name>"). Android names its outputs `speaker`, `earpiece`, `bluetooth`, `wired-headset` and iOS uses port UIDs, so `_platformDefaultIn` matched nothing on a phone and the in-call menu marked no row at all — exactly the case the resolution was added for.

Rather than switch on platform, the controller now follows the call: `forCall` passes a `devicesInUse` stream mapped off `call.state`, since `Call` already resolves a device per kind as it joins (speaker vs. earpiece from the call settings, with the iOS external-device carve-out) and records every switch after that, including the ones iOS makes through its own route picker. Those arrive assigned rather than selected — the call already switched, so there is nothing to apply.

The `default` lookup stays as a fallback: on web with `cameraDefaultOn`, `_applyCallSettingsToConnectOptions` finds no `speaker` id and resolves no output, so the state names no device and the browser default is all there is to mark.

Tests cover the native shape (real ids, no `default` entry) driving a real `CallState`. The lobby is unaffected — it has `supportsSystemDefault: true`, so it marks a real "System default" row and hands the choice to the call at join.

PR #1332, opened 2026-09-09
feat(ui): pick any participant layout from a menu

renefloor:
`StreamLayoutButton` flipped between two layouts and drew one of two icons. It now draws the layout in effect and offers the rest through a `StreamAdaptiveMenuAnchor`: a context menu on a desktop platform, a bottom sheet on Android and iOS. Given two layouts it toggles instead of opening a menu, which is what its default pair does.

Design: [menu](https://www.figma.com/design/W7pOBtNINkAh14XdCJQKuB/Video-SDK-Design?node-id=782-71821), [sheet](https://www.figma.com/design/W7pOBtNINkAh14XdCJQKuB/Video-SDK-Design?node-id=777-29351).

## The seven layouts

| Mode | Label | Renders |
|---|---|---|
| `auto` | Default | Speaker full-frame in a one-on-one call, a grid otherwise. The self-view floats over the speaker, and over the grid on mobile while at most two other people are in the call. |
| `grid` | Grid | Even grid, local participant as an ordinary tile. |
| `speakerTop` / `speakerBottom` / `speakerLeft` / `speakerRight` | Speaker (…) | Spotlight on that edge, with the participants bar along the opposite one. |
| `speakerOneToOne` | Speaker 1:1 | Spotlight alone, local participant floating over it. |

**A layout is named after where the speaker sits, not the bar** — that is what the design's icons draw, so `speakerTop` sets `ParticipantsBarAlignment.bottom`. The four map onto the alignment `CallParticipantsSpotlightView` already had and `RegularCallParticipantsContent` never set. Screen share is untouched: `StreamCallParticipants.build` returns `ScreenShareCallParticipantsContent` before `layoutMode` is read, and that widget picks its own alignment from the window aspect ratio.

## Notes for review

- **`auto` is the new default** for `StreamCallContent`, `StreamCallParticipants` and `RegularCallParticipantsContent`. `grid` now gives the local participant a tile instead of floating them. The livestream widgets still default to `grid`, and are otherwise untouched — the new layouts there are for a later PR with designs of their own.
- **`spotlight` → `speakerTop`, `pictureInPicture` → `speakerOneToOne`**, both deprecated with `fix_data.yaml` transforms. `spotlight` took the spotlight view's default bottom bar, which is `speakerTop` under the naming above. `dart fix --apply` migrates call sites; verified end to end on a probe file.
- **`enableLocalVideo` is `enableFloatingSelfView` now**, and `localVideoParticipantBuilder` is `floatingSelfViewBuilder`. The flag decides whether the self-view *floats over* the layout or the local participant takes a tile like anybody else — the old name read as "show my video at all", which the grid and the four bar layouts ignore because those give them a tile either way. `dart fix --apply` renames both.
- **Only `auto` and `speakerOneToOne` read it**, the two layouts that leave the local participant out of the arrangement. Under `speakerOneToOne` it defaults to on; under `auto` in a group call, to on for mobile with at most two other people and off beyond that. That threshold matches the other SDKs: Android gates on `callParticipants.size in 2..3`, React Native on `remoteParticipants.length < 3`, Swift on `participants.count <= 3`.
- **`speakerOneToOne` floats the self-view on desktop too**, where the flag otherwise defaults to off. The inset *is* that layout — without it the local participant, who gets no tile and no bar either, disappears from the call. An explicit `false` still wins.
- **A speaker layout needs somebody to spotlight.** `participants.first` threw `StateError` on an empty list; five layouts reach that branch now and each is one tap away, so it falls through to the grid instead.
- **A layout with no bar always spotlights a remote.** With the self-view off, the speaker sort could put the local participant first and spotlight them, and since that layout draws no bar, every remote fell off the screen.
- **Changing the layout re-sorts.** `StreamCallParticipants` derives its sort from `layoutMode` but recalculated only when the participants or the call changed, so picking a speaker layout swapped the preset without reordering and spotlighted whoever the previous preset had first. On a quiet call nothing arrived to correct it. Pre-existing, but a seven-entry menu makes it reachable.
- **`StreamContextMenuAnchor.defaultActionStyle` is new** and is the interesting bit. The design fills the selected row, which the sheet did via `StreamListTile(selected:)` but the anchored menu could not — `StreamContextMenuAction` has no selected state, and the radio indicator that would otherwise mark it is suppressed by a per-row `leading` icon. Nesting a second `StreamContextMenuActionTheme` does *not* work: `.of` uses `dependOnInheritedWidgetOfExactType`, so it shadows rather than layers, and the selected row would drop to core's 40px/242px/20px defaults while its siblings keep the design's 32px/200px/16px. Hoisting the row style makes it re-mergeable. A test pins the metrics alongside the fill.
- **`menuDirection`, not `direction`**, matching the device split buttons, and defaulting to `down` alongside them so every call control with a menu defaults the same way.
- The sample offers every layout, minus `speakerLeft`/`speakerRight` below 768px, where a column of participants beside the speaker leaves the speaker too narrow.

## Testing

`flutter analyze` clean for `stream_video_flutter` and the sample. 399 tests pass locally; the 38 golden tests need CI to generate their baselines.

Four test files cover the menu, the toggle, `menuDirection`, what each layout renders, the self-view rules including the mobile threshold and the desktop default, the re-sort on a layout change, and the enum's own derivations — `canonical`, `selectable`, `barAlignment`, `isSpeakerLayout` and `sorting`. The two fixes above each have a test that fails without the change. The menu is asserted rather than snapshotted — the CI golden capture drops anything painted into an `Overlay`.

Checked by eye on macOS and iOS against the Figma: seven rows on the menu, five on the sheet at phone width, per-row icons, active row filled.

### DIFF
--- a/packages/stream_video_flutter/lib/src/call_participants/call_participants.dart
+++ b/packages/stream_video_flutter/lib/src/call_participants/call_participants.dart
@@ -49,9 +49,9 @@ class StreamCallParticipants extends StatefulWidget {
     this.participants,
     this.filter = _defaultFilter,
     CallParticipantSort<CallParticipantState>? sort,
-    this.enableLocalVideo,
+    this.enableFloatingSelfView,
     this.callParticipantBuilder = _defaultParticipantBuilder,
-    this.localVideoParticipantBuilder,
+    this.floatingSelfViewBuilder,
     this.screenShareContentBuilder,
     this.screenShareParticipantBuilder = _defaultParticipantBuilder,
     this.layoutMode = ParticipantLayoutMode.auto,
@@ -74,15 +74,19 @@ class StreamCallParticipants extends StatefulWidget {
   ///
   /// Only [ParticipantLayoutMode.auto] and
   /// [ParticipantLayoutMode.speakerOneToOne] read it, the layouts that leave
-  /// the local participant out of the arrangement. Defaults to true under
-  /// `speakerOneToOne`, and to false on desktop under `auto`.
-  final bool? enableLocalVideo;
+  /// the local participant out of the arrangement. The grid and the four bar
+  /// layouts give them a tile, so a self-view would show them twice.
+  ///
+  /// Defaults to true under `speakerOneToOne`, which `auto` resolves to in a
+  /// one-on-one call. Under `auto` in a group call it defaults to true on
+  /// mobile and false on desktop.
+  final bool? enableFloatingSelfView;
 
   /// Builder function used to build a participant grid item.
   final CallParticipantBuilder callParticipantBuilder;
 
-  /// Builder function used to build a local video participant widget.
-  final CallParticipantBuilder? localVideoParticipantBuilder;
+  /// Builder function used to build the floating self-view.
+  final CallParticipantBuilder? floatingSelfViewBuilder;
 
   /// Builder function used to build a screen sharing item.
   final ScreenShareContentBuilder? screenShareContentBuilder;
@@ -184,9 +188,9 @@ class _StreamCallParticipantsState extends State<StreamCallParticipants>
       call: widget.call,
       participants: sortedParticipants,
       layoutMode: widget.layoutMode,
-      enableLocalVideo: widget.enableLocalVideo,
+      enableFloatingSelfView: widget.enableFloatingSelfView,
       callParticipantBuilder: widget.callParticipantBuilder,
-      localVideoParticipantBuilder: widget.localVideoParticipantBuilder,
+      floatingSelfViewBuilder: widget.floatingSelfViewBuilder,
     );
   }
 }
--- a/packages/stream_video_flutter/lib/src/call_participants/call_participants.dart
+++ b/packages/stream_video_flutter/lib/src/call_participants/call_participants.dart
@@ -70,7 +70,12 @@ class StreamCallParticipants extends StatefulWidget {
   /// Used for sorting the call participants.
   final CallParticipantSort<CallParticipantState> sort;
 
-  /// Enable local video view for the local participant.
+  /// Whether the local participant's self-view floats over the layout.
+  ///
+  /// Only [ParticipantLayoutMode.auto] and
+  /// [ParticipantLayoutMode.speakerOneToOne] read it, the layouts that leave
+  /// the local participant out of the arrangement. Defaults to true under
+  /// `speakerOneToOne`, and to false on desktop under `auto`.
   final bool? enableLocalVideo;
 
   /// Builder function used to build a participant grid item.
--- a/packages/stream_video_flutter/lib/src/call_participants/call_participants.dart
+++ b/packages/stream_video_flutter/lib/src/call_participants/call_participants.dart
@@ -79,7 +79,8 @@ class StreamCallParticipants extends StatefulWidget {
   ///
   /// Defaults to true under `speakerOneToOne`, which `auto` resolves to in a
   /// one-on-one call. Under `auto` in a group call it defaults to true on
-  /// mobile and false on desktop.
+  /// mobile while at most two other people are in the call, and to false
+  /// otherwise.
   final bool? enableFloatingSelfView;
 
   /// Builder function used to build a participant grid item.
--- a/packages/stream_video_flutter/lib/src/call_participants/call_participants.dart
+++ b/packages/stream_video_flutter/lib/src/call_participants/call_participants.dart
@@ -153,13 +153,18 @@ class _StreamCallParticipantsState extends State<StreamCallParticipants>
   void didUpdateWidget(covariant StreamCallParticipants oldWidget) {
     super.didUpdateWidget(oldWidget);
 
+    // The sort follows layoutMode unless the caller supplied one, so picking
+    // a speaker layout swaps the preset and the list has to be ordered again.
+    final sortChanged = widget.sort != oldWidget.sort;
+
     if (widget.participants != null) {
       _participantsSubscription?.cancel();
 
-      if (!const ListEquality<CallParticipantState>().equals(
-        widget.participants!.toList(),
-        oldWidget.participants?.toList(),
-      )) {
+      if (sortChanged ||
+          !const ListEquality<CallParticipantState>().equals(
+            widget.participants!.toList(),
+            oldWidget.participants?.toList(),
+          )) {
         recalculateParticipants(widget.participants!);
       }
     } else if (widget.call != oldWidget.call) {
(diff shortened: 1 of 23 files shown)


packages/stream_video_flutter/lib/src/call_participants/regular_call_participants_content.dart
```diff
@@ -54,44 +70,86 @@ class RegularCallParticipantsContent extends StatelessWidget {
...
+    // arrangement, so only they can float a self-view. An explicitly
+    // requested grid and the four bar layouts give them a tile, where a
+    // self-view would show them twice, so every other case is false whatever
+    // enableFloatingSelfView says. auto's grid still floats one.
+    //
+    // speakerOneToOne floats on every platform: it shows nobody but the
+    // speaker, so without the inset the local participant is absent from the
+    // call entirely. auto's grid follows the platform and the size of the
+    // call instead.
+    final floatsSelfView = switch (effective) {
```
Brazol: The expansion walks the migration step by step rather than just asserting the flag is ignored: grid was the old default, the old code floated the self-view on mobile for it, so "grid + self-view" was the common configuration — then dart fix renames the parameter, the code compiles, nothing warns, and the self-view is gone. It's explicit that the design looks right (a grid tile plus a floating copy would draw the local participant twice) and that only the silence is the problem. I also flagged a wrinkle in my own suggested assert: effective depends on participant count under auto, so a naive assert would fire when a 1:1 call grows to three — gating on layoutMode instead avoids that. Ends with a proposed CHANGELOG line and the enableLocalVideo → enableFloatingSelfView typo in the existing Breaking entry.

packages/stream_video_flutter/lib/src/call_participants/regular_call_participants_content.dart
```diff
@@ -54,44 +70,86 @@ class RegularCallParticipantsContent extends StatelessWidget {
...
+      ParticipantLayoutMode.speakerOneToOne => enableFloatingSelfView ?? true,
+      ParticipantLayoutMode.grid when isAuto =>
+        enableFloatingSelfView ??
+            (!isDesktopDevice &&
+                remoteParticipants.length <= _maxRemotesBehindSelfView),
+      _ => false,
```
Brazol, line 101: The `_ => false` arm means `enableFloatingSelfView` is read by two of the seven layouts. For `grid`, the four bar layouts and both deprecated aliases, whatever the caller passes is discarded — no assert, no warning, no analyzer hint.

The docs on both `enableFloatingSelfView` fields state this plainly, and I don't think the design is wrong: a grid already gives the local participant a tile, so floating a self-view on top would draw them twice. My worry is only that the ignored case is invisible at runtime, and the upgrade path walks people into it.

**The upgrade path.** `grid` used to be the default `layoutMode`, and a grid floated the self-view on mobile:

```dart
// before
final enableLocalVideo = this.enableLocalVideo ?? !isDesktopDevice;
final showLocalVideo = enableLocalVideo && remoteParticipants.isNotEmpty;

File: packages/stream_video_flutter/lib/src/call_participants/call_participants.dart [hunk 3]
```diff
@@ -144,13 +154,18 @@ class _StreamCallParticipantsState extends State<StreamCallParticipants>
   void didUpdateWidget(covariant StreamCallParticipants oldWidget) {
     super.didUpdateWidget(oldWidget);
 
+    // The sort follows layoutMode unless the caller supplied one, so picking
+    // a speaker layout swaps the preset and the list has to be ordered again.
+    final sortChanged = widget.sort != oldWidget.sort;
```
Brazol: The behaviour here is right: `sort` defaults to `layoutMode.sorting`, so picking a speaker layout swaps the preset, and without a re-sort the spotlight would keep showing whoever the previous layout put first. `call_participants_sort_test.dart` pins exactly that. No argument with the fix — just with detecting it by comparing `CallParticipantSort` values, since that's function identity.

**Why it works today.** `CallParticipantSortingPresets.regular` and `.speaker` are `static final`, so each is a single instance created once. `regular != speaker` is true, `regular != regular` is false, and the comparison does the right thing for anyone who leaves `sort` null.

**Where it breaks.** `CallParticipantSort` is a function type, so `!=` is identity for anything that isn't one of those cached instances. An inline closure is a fresh object every build:

```dart
StreamCallParticipants(
  call: call,
  sort: (a, b) => a.name.compareTo(b.name),  // new object each build
)
```

packages/stream_video_flutter/lib/src/call_participants/regular_call_participants_content.dart
renefloor: This one is deliberate. We don't want a floating self-view on the layouts that already give the local participant a tile, and I'd rather not spam devs with logs or asserts for it — leaving `enableFloatingSelfView` on `true` while switching layouts with the button is fine, `auto` and `speakerOneToOne` honour it and the rest give them a tile.

You're right that the upgrade path deserved to be louder in the changelog though. The breaking entry now says the flag is *ignored* rather than merely unread:

> `enableFloatingSelfView` is read only by `auto` and `speakerOneToOne`; on `grid` and the four bar layouts the local participant takes a tile and an explicit `true` is ignored.

packages/stream_video_flutter/lib/src/call_participants/call_participants.dart
renefloor: Good catch, and it turned out to be worse than latent — I checked, and even a non-capturing closure compares unequal across builds:

```
non-capturing closure:            sort != sort -> true
capturing closure (same value):   sort != sort -> true
preset, same mode:                sort != sort -> false
preset, mode changed:             sort != sort -> true
```

So any call site passing `sort:` inline was re-sorting on every parent rebuild, not just risking it.

Fixed in d7cd8e7c by comparing through `layoutMode` instead, so both sides are always one of the two cached presets and a caller's closure never enters the comparison:

```dart
final sortChanged =
    widget.layoutMode.sorting != oldWidget.layoutMode.sorting;
```

It also reads as what the bug actually was: the layout changed the preset. There's a test with a counting comparator handed over fresh each build — three pumps report 3 sorts with the old comparison and 1 with this one.

I've deliberately left the remaining case, where a caller changes their own `sort` without the participants changing. Closing it properly means keeping `sort` nullable on the widget and resolving it in the state, and it buys little: on the default path `partialState` only dedupes by element equality, and `audioLevel` / `audioLevels` / `isSpeaking` are all in `CallParticipantState.props`, so any speech re-emits the list and the new comparator lands within a frame or two. The quiet call is the case that actually needed fixing, and `layoutMode` covers it.

Brazol (approved):

PR #1337, opened 2026-09-10
chore(repo): network layer cleanup

Brazol:
## Summary by CodeRabbit

* **Breaking Changes**
  * Renamed public error types to `StreamVideoException` and `StreamVideoExceptionWithCause`; previous names are deprecated.
  * Removed `RetryConfig.callRejoinTimeout` and added separate retry controls for total delay and SFU operations.

* **Improvements**
  * Improved request retries, token refresh handling, `Retry-After` support, and error reporting.
  * Calls now handle non-reconnectable SFU failures more reliably and remain disconnected after remote call termination.
  * Requests can proceed while socket connection details are unavailable.

* **Diagnostics**
  * Added tracking for dropped or malformed coordinator messages.


Brazol:
@coderabbitai review

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -562,34 +561,36 @@ class StreamVideo extends Disposable {
...
+
+    _logger.d(() => '[rewatchCalls] count: ${watched.length}');
+    unawaited(
+      queryCalls(
+        watch: true,
+        filterConditions: {
+          'cid': {r'$in': watched.map((call) => call.callCid.value).toList()},
+        },
+      ).onError((error, stackTrace) {
+        _logger.e(() => '[rewatchCalls] re-watching calls failed: $error');
```
renefloor: **This error path is dead code.** `queryCalls` wraps its body in `try/catch` and *returns* `Result.failure` - it never throws - so `Future.onError` never fires. A failed re-watch produces no log at any level and no state change.

That matters more after this PR than before it: with `_waitUntilConnected()` gone, `_resolveConnectionId`'s doc comment names `_rewatchCalls` as the mechanism that repairs watches registered without a connection id. If a reconnect's re-watch is answered with a 429 or a token refusal, the call stays in `watchedCalls`, receives no events, and nothing retries until the *next* `CoordinatorConnectedEvent` - so the ringing-call fix in the changelog silently doesn't happen.

Fold the result instead:
```dart
unawaited(queryCalls(...).then((result) {
  if (result case Failure(:final error)) {
    _logger.e(() => '[rewatchCalls] re-watching ${watched.length} call(s) failed: $error');
  }
}));
```

renefloor (commented):
Reviewed the full diff against `v2`, plus the deleted code and the relevant `stream_core` 0.5.0 / `dio` sources. CI is green and the 652 tests pass locally.

This is a solid refactor and I want to be clear about that before the list: collapsing the 43 per-method `_waitUntilConnected()` guards into a lazy connection-id interceptor is a large, real simplification, and `StreamRetryInterceptor._isRetryable` - an exhaustive switch over a sealed hierarchy with a written justification per arm and no default case - is the best type work in the PR. Adding a fifth kind in core becomes a compile error here rather than a silent fall-through to "retry". The auth path in `coordinator_ws.dart` logs at error level before every terminal throw and carries the original error through `Error.throwWithStackTrace` instead of flattening it. And `_CountingAdapter` in `stream_retry_interceptor_test.dart` is a fake deliberately built not to lie about retry state.

The findings cluster in error *reporting* rather than error *logic*. Two I would treat as blocking:

- **`_rewatchCalls`'s error path is dead code** - `queryCalls` returns `Result.failure` rather than throwing, so `.onError` never fires. This is the mechanism that compensates for removing `_waitUntilConnected()`, and it swallows its own failure silently.
- **Token refresh failures are discarded with no logging** - and `stream_auth_interceptor.dart` has no logger at all, so a customer's failing token backend is unattributable.

Then `_awaitConnected` reporting auth refusals as timeouts, an unreachable retry branch, and `SfuClient` retrying refusals.

On testing, the deleted-test audit comes out well: 4 of the 5 removed files are fully re-covered (`stream_video_exception_composer_test.dart` is a strict superset of the old composer test, and `ws_test.dart`'s coverage legitimately moved into `stream_core`). The gaps are `SfuWebSocket._handleDisconnected`, whose branches this PR *inverts* with nothing in the suite constructing an `SfuWebSocket`, and the socket half of the static-provider token guard in `coordinator_ws.dart:156`.

One thing I checked and would explicitly **not** change: the `unrecoverable: false` handling. It looks like a regression against `v2`'s `if (unrecoverable != null) return !unrecoverable;`, but the test at `stream_retry_interceptor_test.dart:111` documents that the backend declares the field `omitempty` - a `false` never reaches the wire, so both versions fall through to the status heuristic identically. Deliberate, documented, tested.

Details inline.

packages/stream_video/lib/src/http/stream_auth_interceptor.dart
```diff
@@ -0,0 +1,114 @@
...
+
+    // A provider that always returns the same token has nothing fresher to
+    // give — a guest's access token included.
+    if (tokenSource.usesStaticProvider) return handler.next(err);
+
+    // The replacement was refused too, so another one will be as well.
+    if (options.extra[_retriedKey] == true) return handler.next(err);
+
+    final refreshed = await tokenSource.refreshToken();
+    if (refreshed is! Success<UserToken>) return handler.next(err);
```
renefloor: **The refresh failure is discarded silently** - and this file has no logger at all, so nothing about it is recoverable afterwards.

This `Failure` is where an app's own `tokenProvider` exception lands (composed in `token_manager_extension.dart`). What the caller sees instead is the original 401. If a customer's token backend starts returning 500s, every request reports "token expired" forever and the one fact that explains it never reaches logs or Sentry.

Worth logging at error level, and ideally attaching the refresh failure as the cause on the rejected exception rather than passing the original refusal through untouched.

packages/stream_video/lib/src/coordinator/open_api/coordinator_client_open_api.dart
```diff
@@ -134,41 +161,72 @@ class CoordinatorClientOpenApi extends CoordinatorClient {
...
-        .firstWhere(
-          (it) => it.isConnected,
-        )
-        // TODO
-        // replace timeout with config value,
-        .timeout(const Duration(milliseconds: _waitForConnectionTimeout))
-        .then((it) {
-          _logger.v(() => '[waitUntilConnected] completed: $it');
-          return const Result.success(none);
+        .firstWhere((it) => it.isConnected)
```
renefloor: **A terminal auth failure surfaces here as a timeout.** `firstWhere((it) => it.isConnected)` never matches the `disconnected(apiError: ...)` state that `connectUser`'s subscription sets on an auth refusal, so this waits out the full `_connectUserTimeout` and fails with a bare `TimeoutException` - discarding the verdict (`token_signature_invalid`, `api_key_invalid`, `unrecoverable: true`) that is sitting in `_connectionState.value`.

That contradicts the doc comment just above ("a caller awaiting it needs the failure") - it gets *a* failure, not *the* failure. Practical cost: an integrator with a wrong API key gets a timeout, and any retry-on-timeout logic loops forever against a config error. Note `StreamRetryInterceptor` classifies these same conditions as terminal, so the two paths disagree.

Suggest waiting on either terminal state (`isConnected || isDisconnected`) and reading `apiError` off the disconnected one.

packages/stream_video/lib/src/http/stream_retry_interceptor.dart
```diff
@@ -0,0 +1,157 @@
...
+    // the cancellation.
+    StreamNetworkException(isCancelled: true) => false,
+
+    // Never reached a verdict — a timeout, a dropped connection. Worth another
+    // attempt, and the outcome of the last one is unknown either way.
+    StreamNetworkException() => true,
+
+    // Credentials that could not be produced or sent. Retried because the
+    // failure may be a transient one inside app-supplied token code.
+    StreamAuthenticationException() => true,
```
renefloor: **This branch looks unreachable.** The only producer of a `StreamAuthenticationException` in this chain is `StreamAuthInterceptor.onRequest`'s `handler.reject(...)`, and dio's `RequestInterceptorHandler.reject(error, [callFollowingErrorInterceptor = false])` defaults to skipping the following error interceptors. That call doesn't pass `true`, and `StreamCoreHttpClient` adds no interceptors of its own - so `onError` here is never invoked for a signing failure, and the retry the comment describes never happens.

Either pass `true` at the reject site or drop the arm and its comment. Also worth noting no test puts the auth and retry interceptors in the same chain - the two are only ever exercised apart, which is what let this through.

packages/stream_video/lib/src/sfu/sfu_client.dart
```diff
@@ -68,48 +75,57 @@ class SfuClient {
...
-          await Future<void>.delayed(
-            Duration(milliseconds: _retryInterval(attempt)),
-          );
-          continue;
-        }
-        return Result.failure(VideoErrors.compose(e, stk));
+        _logger.w(
+          () => '[$label] SFU asked to retry: ${sfuError.message}',
+        );
+      } on Exception catch (e, stk) {
```
renefloor, line 105: Widening `on TimeoutException` to `on Exception` means refusals are now retried. `TwirpError implements Exception`, so an SFU RPC refused with `permission_denied` / `unauthenticated` / `invalid_argument` now burns all `sfuRpcMaxRetries` attempts with backoff before failing - roughly 0.6s of dead air on a live negotiation, and that latency sits inside the reconnect budget when it happens during `sendAnswer`.

This contradicts the philosophy encoded ~40 lines away in `StreamRetryInterceptor._isRetryable` ("a retry reproduces the same configuration, so it cannot land differently"). Consider classifying, so a 4xx-equivalent Twirp code fails on the first attempt.

Separately: this loop doesn't consult `maxTotalRetryDelay` at all, so that one config field has two enforcers and only one obeys it.

packages/stream_video/lib/src/sfu/ws/sfu_ws.dart
```diff
@@ -129,7 +130,7 @@ class SfuWebSocket {
   Future<Result<None>> recreate() async {
     _logger.i(() => '[recreate] no args');
     await _client.disconnect(
-      closeCode: CloseCode(StreamWebSocketCloseCode.disposeOldSocket.value),
+      closeCode: StreamVideoCloseCode.disposeOldSocket,
```
renefloor, line 133: `.catchError((_) => _client.connectionState.value)` is an empty catch with a return value. If the old socket doesn't reach `Disconnected` in 10s - or the emitter completes without one, giving a `StateError` - nothing is logged, the code proceeds to `connect()` on a client still in the old state, and still returns `Result.success(none)`.

The caller trusts that: `call_session.dart:508-513` checks `isFailure`, sees success, and sends the join request into a socket that may not be ready (see the `send()` guard below). Worth branching on the timeout and returning a failure.

packages/stream_video/lib/src/sfu/ws/sfu_ws.dart
```diff
@@ -167,6 +168,16 @@ class SfuWebSocket {
 
   void send(sfu_events.SfuRequest message) {
     _logger.v(() => '[send] message: $message');
+
+    // `leave` and `sendPing` both fire opportunistically, so they can reach a
+    // socket that was never opened — a session abandoned before it connected.
+    // The client reports sending before connecting as a [StateError], which is
+    // nothing for those callers to act on: there is no connection to send on.
+    if (_client.connectionState.value is Initialized) {
```
renefloor, line 172: The comment says `leave` and `sendPing` "both fire opportunistically" - but `send()` has four callers and two of them are the JOIN request (`call_session.dart:315` in `start`, `:517` in `fastReconnect`). Neither is opportunistic; both immediately `await sfuWS.events.waitFor<SfuJoinResponseEvent>()`.

So a join dropped here isn't a suppressed no-op - it's a 30s block on a response to a request that was never sent, surfacing as a generic `TimeoutException`, with the only explanation being a `w` log in another file.

Suggest returning a `Result` from `send()` so the join sites can fail fast, and having `leave`/`sendPing` ignore it explicitly.

packages/stream_video/lib/src/sfu/ws/sfu_ws.dart
```diff
@@ -190,23 +201,29 @@ class SfuWebSocket {
   void _handleDisconnected(DisconnectionSource source) {
     switch (source) {
       case ServerInitiated(:final error?):
-        if (error.error != null) {
+        // A bare closure — the socket closed with a code and a reason, with no
+        // error underneath it — is a plain disconnection. Everything else
+        // failed the connection: a server verdict, a transport error the
+        // closure wraps, or an SDK-side failure.
+        if (error case StreamNetworkException(cause: null, :final closeCode)) {
```
renefloor, line 204: This condition is inverted relative to `v2` (`error.error != null` -> `error case StreamNetworkException(cause: null, ...)`), so `ServerInitiated` now emits `SfuSocketDisconnected` where it previously emitted `SfuSocketFailed` and vice versa - and nothing in the suite constructs an `SfuWebSocket` (`grep -rln "SfuWebSocket(" test/` is empty).

The two new adjacent tests bracket this without touching it: `call_sfu_socket_closure_test.dart` tests the consumer with `isReconnectable` already set, and `sfu_reconnect_policy_test.dart` pins core's `DisconnectionSource`. The producer in between - which decides whether a call reconnects or dies - is the part this PR actually changed.

The pattern already exists: `coordinator_ws_disconnect_test.dart:37-42` drives `CoordinatorWebSocket` through `@visibleForTesting client`, and `SfuWebSocket` exposes the same hook. Worth covering: bare `StreamNetworkException(cause: null)` -> `Disconnected`; one with a `cause` -> `Failed`; a `StreamApiException` verdict -> `Failed`; `closeCode: 0` -> `null`.

Also: a `ServerInitiated` closure with a **null** error matches neither this arm nor any other, falling to the final `break` - emitting nothing at all. Intended?

packages/stream_video/lib/src/sfu/ws/sfu_ws.dart
```diff
@@ -190,23 +201,29 @@ class SfuWebSocket {
...
-            SfuSocketDisconnected(
+            SfuSocketFailed(
               sessionId: sessionId,
               url: url,
-              reason: DisconnectionReason(
-                closeCode: error.code != 0 ? error.code : null,
-                closeReason: error.reason != 'Unknown' ? error.reason : null,
-              ),
+              isReconnectable: source.isReconnectable,
+              error: StreamVideoExceptions.compose(error.cause ?? error),
```
renefloor: `error.cause ?? error` throws away the classification. `error` is already a `StreamException` (`isTimeout`, `isCancelled`, or a `StreamApiException` with a status); unwrapping to `.cause` replaces it with the raw `SocketException`/`WebSocketChannelException`, and `compose` wraps that instead.

Downstream, `StreamVideoException.streamException` then answers `null`, so `ClientEventErrorMapper.map` never reaches `_fromStreamException` and falls back to the `_looksOffline` string heuristic - a transport timeout gets reported as `serverError` rather than `requestTimeout`, misattributing a client-side network problem to the SFU in exactly the telemetry this PR builds.

`compose` already handles `StreamException` and keeps it as the cause, so `compose(error)` should do the right thing.

packages/stream_video/lib/src/call/call.dart
```diff
@@ -1445,7 +1456,7 @@ class Call {
       if (disconnectOnMaxRetries) {
         await leave(
           reason: DisconnectReason.failure(
-            VideoError(
+            StreamVideoException(
```
renefloor: The retry loop above holds `error` on every iteration but never keeps it, so this failure reaches the app with no cause at all - `apiStatusCode`, `apiErrorCode`, `isUnrecoverable`, `retryAfter` and `sfuError` all answer `null`/`false`.

That's the single failure an integrator is most likely to handle, and every accessor this PR introduces returns nothing on it. Three attempts against a 429 with `Retry-After: 60` are indistinguishable from a device with no network - the two cases needing opposite UI.

Carrying the last error into both the `leave` reason and the returned failure would fix it.

packages/stream_video/lib/src/coordinator/open_api/coordinator_ws.dart
```diff
@@ -122,42 +134,56 @@ class CoordinatorWebSocket {
...
+      final StreamApiException it when it.isTokenExpired => it,
+      _ => null,
+    };
+    final tokenRefused = refusal != null;
 
     // Mirrors the RpcRetryManager guard: a static provider can only return
     // the token the server just refused, so the credentials cannot change.
-    // Throwing fails the attempt for good (AuthenticationFailed) instead of
-    // reconnecting with the same dead token.
+    // Throwing fails the attempt for good (AuthenticationFailed, which is not
```
renefloor, line 154: `RpcRetryManager` is deleted by this PR (`lib/src/retry/retry_manager.dart`), and this is the only remaining mention in `lib/` - it now points at a class that can't be found. The rest of the comment stands on its own, so dropping the first clause is enough.

(Same name also lingers in `test/src/token/build_token_provider_test.dart:179`.)

packages/stream_video/lib/src/call/call.dart
```diff
@@ -4151,7 +4187,7 @@ class Call {
       _stateManager.participantSetAudioInputDevice(device: device);
       return const Result.success(none);
     } else {
-      if (result.getErrorOrNull() case VideoErrorWithCause(
+      if (result.getErrorOrNull() case StreamVideoExceptionWithCause(
```
renefloor: This is the only remaining use in `lib/` of the `cause` getter this PR deprecates. `rawCause` was added at `stream_video_exception.dart:127` and documented as being "for the handful of internal types the SDK itself puts here and reads back" - which is exactly `TrackMissingException`. Switching keeps the deprecation honest.

packages/stream_video/lib/src/coordinator/open_api/coordinator_message_codec.dart
```diff
@@ -42,11 +32,71 @@ final class CoordinatorWsEvent extends WsEvent {
   static const suppressed = CoordinatorWsEvent(null);
 }
 
+/// Why a coordinator message was dropped instead of delivered.
+enum CoordinatorDropReason {
```
renefloor: `CoordinatorDropCounter` has no production reader. `codec.dropped` is referenced only in `coordinator_message_codec_test.dart`; the one production instance (`coordinator_ws.dart:55`) is constructed inline and never held, so nothing can ever read the counts.

That undercuts the doc comment's rationale - "without it, 'the event never arrived' cannot be told apart from 'the server never sent it'" - since no support engineer can obtain these counts from a user's device. In production the net change is five `_logger.w` lines where there was one.

Either surface it (`CoordinatorWebSocket` already has `clientEventReporter` right there) or drop ~45 lines of new public API and keep the logging.

Related: `CoordinatorUnknownEvent` is dropped separately at `coordinator_ws.dart:227` at `v` level, uncounted and without naming the type - inconsistent with the accounting here.

packages/stream_video/lib/src/retry/retry_policy.dart
```diff
@@ -42,20 +45,37 @@ class RetryConfig extends Equatable {
...
+  /// How many attempts a coordinator HTTP request gets in total, the first one
+  /// included.
   final int rpcMaxRetries;
 
+  /// The ceiling on the computed backoff between attempts.
   final Duration maxBackoff;
 
-  @Deprecated('Use CallPreferences.networkAvailabilityTimeout instead')
-  final Duration callRejoinTimeout;
+  /// The total time retrying may add to one call.
```
renefloor: "The total time retrying may add to one call" overstates the reach, and "call" reads as the `Call` domain type. The budget is per HTTP request (tracked in `RequestOptions.extra` by `StreamRetryInterceptor`), and `SfuClient._executeWithRetry` doesn't consult it at all.

Suggest narrowing to something like: "The total time retrying may add to a single coordinator HTTP request. Not applied to SFU signalling retries."

While here - `RetryConfig` has no construction-time validation. `rpcMaxRetries: -1`, a negative `maxBackoff`, and `maxTotalRetryDelay < maxBackoff` (a budget smaller than one wait, so retries never happen) are all constructible. It's a `const` constructor, so asserts are free at runtime and would turn three silent misconfigurations into loud ones - and would remove the need for the synthesized-message branch at `sfu_client.dart:117-121`.

packages/stream_video/CHANGELOG.md
```diff
@@ -35,6 +36,13 @@
...
   - `onTokenUpdated` for a static token now fires on first token use and after every refresh, instead of once at client construction.
 - A coordinator WebSocket connection whose token could not be loaded is now closed as an authentication failure instead of being retried indefinitely. A `tokenLoader` that throws on a reconnect therefore ends the connection — the error is reported on `StreamCallDisconnectedEvent`/`CoordinatorDisconnected` — where before the socket kept retrying with nothing reported.
+- `StreamVideoExceptionWithCause.cause` now carries the `StreamApiException` for a failure the server answered, where it previously carried the parsed `StreamApiError` payload. Code matching on `cause is StreamApiError` still compiles but no longer matches, so this change is silent. Read the verdict through the accessors on `StreamVideoException` instead: `apiStatusCode`, `apiErrorCode`, `isUnrecoverable`, `retryAfter`, and `apiError` for the payload itself. They answer from either shape.
```
renefloor, line 39: These entries run well past the one-short-sentence-per-item convention - this one is four sentences, and lines 58-63 each append the mechanism. Line 40 also editorialises about the entry above it ("which is what happened to the change above"). The mechanism reads better in the code comments, where most of it already is.

Two other things in this block:
- Trailing whitespace on this line and on line 45 (the same diff *removes* a trailing space from line 84).
- The ringing entry (line 62) is narrower than the change: `stream_video.dart:573-590` re-watches **every** call in `state.watchedCalls`, not only ringing ones, and moves the trigger from `CoordinatorReconnectedEvent` to `CoordinatorConnectedEvent`. Ringing is the motivating symptom, not the scope.

Also missing: the largest behavioural change here - coordinator calls no longer block on the socket (the 43 `_waitUntilConnected()` guards are gone) - is only mentioned for guest creation.

packages/stream_video/lib/src/http/stream_auth_interceptor.dart
```diff
@@ -0,0 +1,140 @@
...
+
+    final refreshed = await tokenSource.refreshToken();
+    if (refreshed is! Success<UserToken>) {
+      // Reported here as well as by `TokenManager`, because this is the only
+      // place the two halves meet: what the caller receives is the server's
+      // `token expired`, and on its own that reads as an expiry rather than as
+      // an app token endpoint that is failing.
+      //
+      // Not attached to the rejected failure as a cause: the retry interceptor
+      // decides from its type, and a credentials failure there is retryable
```
renefloor: This justification was overtaken by the same commit that wrote it. It says a credentials failure "is retryable" in the retry interceptor, and that swapping it in as the cause would spend five more attempts - but `StreamRetryInterceptor._isRetryable` now classifies `StreamAuthenticationException() => false`, so swapping it in would change nothing.

Not attaching the cause may still be the right call; it is the stated reason that no longer holds. Worth rewording so the next reader doesn't act on it.

While you're in there: that `=> false` arm is now the only one in that switch without a rationale, and its explanation lives only in the test file. A one-liner - unreachable in the real chain because the auth interceptor rejects in `onRequest`, asserted in tests so it cannot drift - would bring it back in line with its neighbours.

renefloor (approved):
Approving.

The two blocking findings are properly closed, and three of the fixes went further than what I raised:

- `_awaitConnected` also fixes a subscription leak I missed - `Future.timeout` completes its own future without cancelling the `firstWhere` underneath, so every request that gave up while the socket was coming up leaked a listener. The new `_awaitConnection` owns the subscription, and correctly declines to settle on transient disconnects, which my suggested fix would have got wrong.
- The unreachable retry branch ships with the wired-together test, plus a deliberate contrast case whose comment says it exists so the first assertion is not passing for the wrong reason.
- `sfu_ws_disconnect_test.dart` covers the inverted branch thoroughly, including the `closeCode: 0` normalisation.

661 tests pass locally, CI green, analyzer unchanged.

One inline note on a comment that this commit left contradicting itself. Everything else outstanding from my last pass is yours to weigh - the ones I would still like to see land at some point, none blocking:

- `sfu_client.dart:105` still retries Twirp refusals three times with backoff via the widened `on Exception`.
- `sfu_ws.dart` `recreate()` still swallows its 10s timeout in `.catchError` and returns `Result.success`.
- `send()` can still drop a JOIN into a 30s `waitFor` hang; the misleading comment went away but the behaviour did not, and the guard now has no explanation at all.
- The CHANGELOG is unchanged: trailing whitespace on lines 39 and 45, the ringing entry describes a narrower scope than the code, and the largest behavioural change here - coordinator calls no longer blocking on the socket - still is not recorded. Worth closing before release, since it is the public record for a breaking version.

PR #1341, opened 2026-09-11
feat(llc): helper for video push handling

Brazol:
## Summary by CodeRabbit

* **New Features**
  * Added streamlined background handling for ringing call notifications, including support when the app is not open.
  * Improved call notification processing across foreground, background, and app-isolate scenarios.
  * Added Android Telecom integration and improved iOS CallKit support for ringing calls.

* **Bug Fixes**
  * Improved ringtone behavior on Android.
  * Improved cleanup and recovery when call notification setup or handling fails.
  * Added support for reporting ended calls on iOS.

### DIFF
--- a/dogfooding/lib/app/app_content.dart
+++ b/dogfooding/lib/app/app_content.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:firebase_messaging/firebase_messaging.dart';
 import 'package:flutter/material.dart';
 import 'package:google_fonts/google_fonts.dart';
@@ -181,7 +183,11 @@ class _StreamDogFoodingAppContentState
   void _observeFcmMessages() {
     FirebaseMessaging.onBackgroundMessage(firebaseMessagingBackgroundHandler);
     _compositeSubscription.add(
-      FirebaseMessaging.onMessage.listen(handleRemoteMessage),
+      FirebaseMessaging.onMessage.listen(
+        (message) => unawaited(
+          StreamVideo.instance.handleRingingFlowNotifications(message.data),
+        ),
+      ),
     );
   }
 
--- a/packages/stream_video_push_notification/CHANGELOG.md
+++ b/packages/stream_video_push_notification/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 ### ✅ Added
 
+- Added `StreamVideoPushHandler.handleBackgroundMessage`, which simplifies the ringing setup by running a Stream ringing push through its whole background lifecycle from your Firebase background handler: it builds the client through the factory you give it, observes the ringing events a background isolate can act on, and disposes the client — along with whatever that factory set up — once the user has answered, declined, or let the call time out. It replaces the setup and teardown each integration had to write by hand. Pass `existingClient` if your app does not keep its client in the `StreamVideo` singleton.
+
+```dart
+@pragma('vm:entry-point')
+Future<void> firebaseMessagingBackgroundHandler(RemoteMessage message) async {
+  await Firebase.initializeApp(options: DefaultFirebaseOptions.currentPlatform);
+
+  await StreamVideoPushHandler.handleBackgroundMessage(
+    message,
+    createStreamVideo: () async {
+      // Read credentials, fetch a token, build the client. Return null when
+      // nobody is logged in.
+    },
+    onDispose: () => MyDependencies.reset(),
+  );
+}
+```
+
 - [Android] Added a Telecom integration for the ringing flow, which registers incoming and outgoing ringing calls with the platform's [Telecom stack](https://developer.android.com/develop/connectivity/telecom) through Jetpack Telecom. This gives the call proper audio focus and a place in the system call state, and lets it be answered or hung up from a paired watch, a car head unit or a Bluetooth headset. The incoming call notification and full-screen ringing UI are unchanged. It is **on by default from Android 17**: for an app targeting API 37 the platform will not play a ringtone from a service started by a push unless the call is in the Telecom stack, so ringing does not work correctly without it. The default follows the Android version of the device rather than your `targetSdk`, so it is on for any app running on Android 17 — if you target below API 37 the restriction does not apply to you and you can opt out with `AndroidPushConfiguration(telecom: TelecomPushConfiguration(enabled: false))`. It is **off by default below Android 17**, where ringing works either way, so an existing integration is unaffected unless you pass `enabled: true`.
 - [iOS] Added `reportCallEnded`, which reports how a call ended to CallKit so it is listed correctly in the system Recents.
 
--- a/packages/stream_video_push_notification/lib/src/background_push_handler.dart
+++ b/packages/stream_video_push_notification/lib/src/background_push_handler.dart
@@ -1,0 +1,302 @@
+import 'dart:async';
+
+import 'package:firebase_messaging/firebase_messaging.dart';
+import 'package:meta/meta.dart';
+import 'package:rxdart/rxdart.dart';
+import 'package:stream_video/stream_video.dart';
+
+/// Builds the client a background push needs, or returns null when there is
+/// nobody to build it for.
+///
+/// Called from a fresh isolate that never ran `main()`, so it has to do
+/// everything from scratch: read stored credentials, fetch a token, construct
+/// [StreamVideo]. Returning null is the normal way to say "no user is logged
+/// in", not an error.
+typedef CreateStreamVideo = Future<StreamVideo?> Function();
+
+/// Returns the client this isolate already has, or null when it has none.
+///
+/// Answers the one question the handler cannot answer for itself: whether it
+/// was called on the app's own isolate, where a client is already running and
+/// its lifecycle belongs to the app.
+///
+/// Defaults to the [StreamVideo] singleton. Pass your own when your app builds
+/// its client with [StreamVideo.create], or holds it anywhere else: such a
+/// client installs no singleton, so there is nothing for the default to find
+/// and the handler would build a second one over the top of it.
+typedef ResolveExistingClient = StreamVideo? Function();
+
+/// Runs a Stream ringing push through its whole background lifecycle.
+///
+/// On Android, Firebase delivers background messages to an isolate that never
+/// ran `main()`, so nothing you set up at launch exists there. Everything
+/// from building the client to tearing it down again needs to happen inside the
+/// handler. On Apple platforms, the same handler is called on the app's own
+/// isolate, where the client already exists and must be left alone—both cases
+/// are handled here.
+///
+/// You keep the parts that are genuinely yours, such as Firebase
+/// initialization and credential management, and hand the rest over:
+///
+/// ```dart
+/// @pragma('vm:entry-point')
+/// Future<void> firebaseMessagingBackgroundHandler(RemoteMessage message) async {
+///   await Firebase.initializeApp(options: DefaultFirebaseOptions.currentPlatform);
+///
+///   await StreamVideoPushHandler.handleBackgroundMessage(
+///     message,
+///     createStreamVideo: () async {
+///       // Read credentials, fetch a token, build the client. Return null if
+///       // nobody is logged in.
+///     },
+///     onDispose: () => MyDependencies.reset(),
+///   );
+/// }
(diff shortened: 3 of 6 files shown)


Brazol:
@coderabbit review

renefloor (commented):
Went through this with a few focused passes (correctness, error handling, tests, comments) and verified the SDK behaviour against source rather than assuming. The design is sound — every issue below is in a failure path, not the shape — and `_releaseSingleton`'s `identical()` guard, the disposed-client check in `_existingClient`, and the ring-vs-missed-call distinction are all genuinely good.

Three critical findings inline. I reproduced the first two with failing tests on this branch (and reverted them); the third is traced through source. There are also a handful of important-but-not-blocking items I've left out of the inline comments — observability of the two `catch` blocks, the cost of building a full auto-connecting client before discovering a push is foreign, `_isRingingPush` duplicating the SDK's payload parsing, and the `ownsInjector` proxy in the dogfooding sample — happy to write those up separately if useful.

Analyzer is clean and all 19 tests pass locally.

packages/stream_video_push_notification/lib/src/background_push_handler.dart
```diff
@@ -0,0 +1,302 @@
...
+  StreamSubscription<RingingEvent>? resolution;
+
+  /// Whether a ringing flow is still waiting to be resolved by the user.
+  bool awaitingResolution = false;
+
+  bool _released = false;
+
+  /// Releases after [delay], so a flow that is still running finishes first.
+  void releaseAfter(Duration delay) =>
+      unawaited(Future<void>.delayed(delay, release));
```
renefloor, line 276: **Critical: this timer can't be cancelled, and tears down a call that started during the grace window.**

No `Timer` handle is kept, and `release()` never re-checks `awaitingResolution`. Both schedulers (`_handleWithSession` L152 and the resolution observer L231) fire and forget, so a teardown armed for one message runs even after a *new* call has started ringing.

Two orderings break. I ran both against this branch and both fail:

1. **Missed call, then a ring 300 ms later.** The missed call arms `releaseAfter(1s)` at L152. The ring sets `awaitingResolution = true`, posts the incoming-call notification, and `handleBackgroundMessage` returns `true`. At t=1 s the orphaned timer still fires: client disposed, observers cancelled, `onDispose` → `AppInjector.reset()`. The user is looking at a ringing screen with nothing behind it, and the decline they tap reaches no client.
2. **Decline, then a redial 300 ms later.** Same shape, armed at L231 instead.

```
REPRO: a ring arriving just after a missed call is not cut short
  Unexpected calls: ... MockStreamVideo.dispose()
REPRO: a second ring during the post-resolution grace is not cut short
  Unexpected calls: ... MockStreamVideo.dispose()
```

The existing test at `background_push_handler_test.dart:220` covers only the ring-then-missed ordering, which is safe precisely because the flag is already set — the reverse is untested.

Suggested fix: hold the pending `Timer` on `_BackgroundSession`, cancel it whenever a message sets `awaitingResolution`, and clear `awaitingResolution` when a resolution event arrives so `release()` can assert on it.

Separately, `unawaited(Future.delayed(...))` means anything `release()` throws becomes an unhandled async error in an isolate with no zone handler — see the comment on L293.

packages/stream_video_push_notification/lib/src/background_push_handler.dart
```diff
@@ -0,0 +1,302 @@
...
+      if (appClient != null) {
+        streamLog.d(
+          _tag,
+          () => '[handleBackgroundMessage] forwarding to the running client',
+        );
+
+        return await appClient.handleRingingFlowNotifications(message.data);
+      }
+
+      final session = await _startSession(createStreamVideo, onDispose);
```
renefloor, line 119: **Critical: two pushes in flight build two sessions; the loser leaks and the winner gets torn down.**

`_session` is read at L106 but only assigned at L235 — after this `await`. There is no in-flight guard, and FlutterFire does not serialise background handler invocations (`method_channel_messaging.dart` installs an `async` method-call handler). Two pushes arriving while the first is awaiting the token fetch both see `_session == null` and both call the factory.

Reproduced with a 50 ms factory and two un-awaited calls: `factoryCalls` was 2, expected 1.

The leak alone is bad — the first session's client is never disposed, its `CompositeSubscription` never cancelled, its `onDispose` never run, so a connected coordinator websocket survives for the life of the isolate. But with the dogfooding factory it cross-wires:

1. B's `AppInjector.registerStreamVideo` hits get_it's "already registered", or `InstanceHolder.install`'s `failIfSingletonExists` throw.
2. B's `_startSession` catch calls `onDispose()` → `AppInjector.reset()`, wiping the DI **sessionA is actively using**, then rethrows.
3. The outer catch at L131 does `await _session?.release()` — and `_session` is now **sessionA**. It disposes A's live ringing client and calls `onDispose` a second time.

Net: two ringing pushes, zero notifications, `onDispose` called twice, and both log lines swallowed.

Suggested fix: store the in-flight `Future<_BackgroundSession?>` in the static slot rather than the resolved session, so a concurrent caller awaits the first setup instead of racing it. And have the outer catch release the session *this* invocation was working with:

```dart
_BackgroundSession? mine;
try {
  ...
  mine = await _startSession(...);
  ...
} catch (e, stk) {
  ...
  await mine?.release();
}
```

packages/stream_video_push_notification/lib/src/background_push_handler.dart
```diff
@@ -0,0 +1,302 @@
...
+    if (_released) return;
+    _released = true;
+
+    if (StreamVideoPushHandler._session == this) {
+      StreamVideoPushHandler._session = null;
+    }
+
+    await resolution?.cancel();
+    await observers.cancel();
+    await streamVideo.dispose();
```
renefloor, line 293: **Critical: teardown isn't failure-safe, and one throw here poisons the isolate permanently.**

`release()` latches `_released = true` (L285) and nulls `_session` (L288) *before* this line, then awaits `dispose()` with no `try`/`finally`. `StreamVideo.dispose` awaits `_client.disconnectUser()`, which — unlike `disconnect()` at `stream_video.dart:500` — is not wrapped in a try/catch.

If it throws:

- `_releaseSingleton` (L298) and `onDispose` (L300) never run, so the singleton slot keeps the client and the app's dependencies stay registered.
- `_released` is already `true`, so nothing can ever retry — every later caller returns at L284.
- `_session` is already `null`, so the outer catch's `await _session?.release()` is a no-op.
- Because `Disposable.dispose()` is `super.dispose()` called **last** inside `StreamVideo.dispose`, `_disposed` was never set — so `isDisposed` is still `false`.

That last point is what makes it permanent rather than merely untidy. On the next push, `_existingClient` (L177) checks `isDisposed`, sees `false`, and forwards to a half-dead client whose subscriptions are cancelled and whose coordinator is disconnected. `handleRingingFlowNotifications` → `getCallRingingState` → `call.get()` fails → folds to `CallRingingState.ended` → returns `false`. Every subsequent incoming call in that isolate is dropped silently.

Suggested fix — make each step best-effort and always reach `onDispose`:

```dart
try {
  await resolution?.cancel();
  await observers.cancel();
  await streamVideo.dispose();
} catch (e, stk) {
  // report
} finally {
  await StreamVideoPushHandler._releaseSingleton(streamVideo);
  await onDispose?.call();
}
```

packages/stream_video_push_notification/lib/src/background_push_handler.dart
```diff
@@ -0,0 +1,450 @@
...
+          return;
+        }
+
+        // The app answered on this client and is in the call, so the client is
+        // the app's now and its lifecycle goes with it. Android delivers the
+        // accept to whichever isolate is running, which is this one whenever
+        // the app is alive but backgrounded - and there the accept is the
+        // moment the app starts using the client, not the moment this session
+        // is done with it. Disposing here drops a call that just connected.
+        if (streamVideo.activeCalls.isNotEmpty) {
```
renefloor: If `activeCalls` is non-empty, the release stands down but never schedules another. When that call ends later, no ringing event arrives to call `releaseAfter` again. The client stays connected, `onDispose` never runs, and `_session` stays set for every later push in this isolate. Could this re-arm when the active call ends, e.g. by listening to `activeCalls` becoming empty?

packages/stream_video_push_notification/lib/src/background_push_handler.dart
```diff
@@ -0,0 +1,450 @@
...
+  }
+
+  /// Handles one Stream push, with this isolate's session to itself.
+  static Future<bool> _handleStreamPush(
+    RemoteMessage message,
+    CreateStreamVideo createStreamVideo,
+    ResolveExistingClient? existingClient,
+    FutureOr<void> Function()? onDispose,
+  ) async {
+    var mine = _session;
```
renefloor: An existing `_session` is reused without checking `streamVideo.isDisposed`, although `_existingClient` does check it. If a session is left behind (see the stand-down in `_release`) and its client gets disposed elsewhere, the next ringing push goes to a dead client and is dropped, so the incoming call never shows. Should a disposed session be released and a fresh one started here?

dogfooding/lib/app/app_content.dart
```diff
@@ -162,6 +170,17 @@ class _StreamDogFoodingAppContentState
     _compositeSubscription.add(
       streamVideo.state.incomingCall.listen((call) {
         if (call == null) return;
+
+        // Only while the app is on screen. In the background the platform
+        // notification is the incoming call UI, and routing here would queue
+        // a screen that is not built until the app resumes - by which time
+        // the call can be long over. Answering from the notification brings
+        // the screen up through `onCallAccepted` instead.
+        if (WidgetsBinding.instance.lifecycleState !=
```
renefloor: `incomingCall` emissions are dropped while the app isn't resumed, and nothing replays them on resume. A call that rings in the background, and is still ringing when the user opens the app from the launcher (or taps the notification body instead of Accept), never gets its incoming-call screen. Could this check `incomingCall.value` again on resume?

dogfooding/lib/screens/call_screen.dart
```diff
@@ -543,7 +562,16 @@ class _CallScreenState extends State<CallScreen>
           onCallDisconnected: (disconnectedProperties) {
             final reason = disconnectedProperties.reason;
 
-            Navigator.of(context).pop();
+            // `pop` closes the topmost route, which is not always this
+            // screen: a dialog can sit above it, and so can a second call
+            // screen. Closing that instead would take down whatever the user
+            // is actually looking at.
+            if (!(ModalRoute.of(context)?.isCurrent ?? false)) {
```
renefloor: When the screen isn't the top route, this returns before the feedback-dialog branch, and the deferred pop in `build` never shows it. So a call that ends with `DisconnectReasonEnded` while a dialog is open closes the screen later without `showFeedbackDialog`.

packages/stream_video_push_notification/CHANGELOG.md
```diff
@@ -18,6 +18,9 @@
...
+- [Android] Added a Telecom integration for the ringing flow, which registers incoming and outgoing ringing calls with the platform's [Telecom stack](https://developer.android.com/develop/connectivity/telecom) through Jetpack Telecom. This gives the call proper audio focus and a place in the system call state, and lets it be answered or hung up from a paired watch, a car head unit or a Bluetooth headset. The incoming call notification and full-screen ringing UI are unchanged. It is **on by default from Android 17**: for an app targeting API 37 the platform will not play a ringtone from a service started by a push unless the call is in the Telecom stack, so ringing does not work correctly without it. The default follows the Android version of the device rather than your `targetSdk`, so it is on for any app running on Android 17 — if you target below API 37 the restriction does not apply to you and you can opt out with `AndroidPushConfiguration(telecom: TelecomPushConfiguration(enabled: false))`. It is **off by default below Android 17**, where ringing works either way, so an existing integration is unaffected unless you pass `enabled: true`.
```
renefloor: This duplicates the Telecom entry already on `v2` a few lines down, and the two describe the default differently. The `reportCallEnded` entry below also looks out of place: that API is already on `v2` and isn't part of this PR. Merge leftover?

renefloor (commented):
A few things from a review pass: two in the background push handler, two in the sample app, one in the changelog.

PR #1343, opened 2026-09-11
fix(samples): match the share call accordion to design

renefloor:
Brings the "Your Meeting is Live!" card in the sample app in line with [Video / iOS / Share Call Accordion](https://www.figma.com/design/W7pOBtNINkAh14XdCJQKuB/Video-SDK-Design?node-id=375-45607&m=dev) ([mobile frame](https://www.figma.com/design/W7pOBtNINkAh14XdCJQKuB/Video-SDK-Design?node-id=373-45309&m=dev), [web frame](https://www.figma.com/design/W7pOBtNINkAh14XdCJQKuB/Video-SDK-Design?node-id=391-22599&m=dev)).

## Spacing

In the design the card sits **8 inside the participant tile** on every side — the same 8 the participant label pill sits at — so their edges line up. We used a flat `EdgeInsets.all(24)`, so nothing did.

The inset is now derived rather than hardcoded: the grid's own padding, read from `StreamCallParticipantsGridTheme`, plus `spacing.xs`. That matters because the sample app overrides the grid padding to 4, so a fixed 16 would have missed the label by 4.

Measured on device against the design:

| | Design | Measured |
|---|---|---|
| Inset inside the tile | 8 left/right/bottom | 8 |
| Collapsed height | 52 | 52 |
| Buttons | 40 + 8 gap + 40 | 40 + 8 + 40 |
| Buttons → QR panel | 16 | 16 |
| QR panel | 160 | 160 |
| Panel → caption | 12 | 12 |
| Caption → card bottom | 16 | 16 |
| Expanded total | 360 | 360 |

## Contents

- Two full-width buttons, **Add Others** and **Copy Call ID**, 8 apart. They pass `tapTargetSize: shrinkWrap` — the default padded target grows each 40 pill to 48 and would double the designed gap; a full-width button keeps a large tap area without it.
- The QR moves onto a `backgroundSurface` panel with its caption underneath instead of above, and the explanatory sentence and the raw call id in the button label are gone.
- Icons come from `context.streamIcons` (`userAddFill`, `copyFill`, `chevronUp`/`chevronDown`, `checkmark`) rather than Material's.
- The card gets the design's drop shadow, through `Material` at `streamElevation.level3` with the border in a foreground `DecoratedBox`.
- A phone gives the card the full width between the insets; a window in the `medium`/`large` bucket centres it at 360, as the web frame draws it.

The "Call ID copied to clipboard" confirmation moves from a Material `SnackBar` to the design system's `StreamSnackbar`, whose `success` variant already carries the checkmark the old one hand-assembled. A `StreamSnackbarScope` sits above the router in `MaterialApp.router`'s builder so any route can reach a messenger. The old bar also rendered its text black on black, from a hardcoded colour that is now gone.

## Verification

Run on the iPhone 18 Pro and iPad (A16) simulators; widget bounds read out of the running app and cross-checked against the Figma geometry — the table above. Copy verified through `simctl pbpaste`. No golden tests cover this widget.

| Before | After | 
| --- | --- |
| before-collapsed | after-collapsed |
| before-expanded | after-expanded |

## Summary by CodeRabbit

- **Bug Fixes**
  - Snackbars shown from any route now remain available and visible while navigating between screens.

- **Style**
  - Updated the Share Call card with refreshed Stream design styling, spacing, colors, borders, and elevation.
  - Improved responsive sizing and layout for the card and participant content.
  - Refined the QR code section with updated sizing, styling, and centered metadata.

- **Usability**
  - Renamed sharing actions to “Add Others” and “Copy Call ID.”
  - Added clearer success feedback when copying a call ID.

### DIFF
--- a/dogfooding/lib/app/app_content.dart
+++ b/dogfooding/lib/app/app_content.dart
@@ -208,7 +208,9 @@ class _StreamDogFoodingAppContentState
       builder: (context, child) {
         return StreamComponentFactory(
           builders: _componentBuilders,
-          child: child!,
+          // Above the router, so a snackbar shown from any route has a
+          // messenger to reach and outlives the screen that showed it.
+          child: StreamSnackbarScope(child: child!),
         );
       },
     );
--- a/dogfooding/lib/widgets/share_call_card.dart
+++ b/dogfooding/lib/widgets/share_call_card.dart
@@ -29,48 +29,84 @@ class _ShareCallWelcomeCardState extends State<ShareCallWelcomeCard> {
 
   @override
   Widget build(BuildContext context) {
-    final theme = StreamVideoTheme.of(context);
-    final colorScheme = StreamTheme.of(context).colorScheme;
+    final colorScheme = context.streamColorScheme;
+    final spacing = context.streamSpacing;
+    final borderRadius = BorderRadius.all(context.streamRadius.lg);
+
+    // The design puts the card's edges on the participant label's. The label
+    // sits spacing.xs inside the tile, and the tile sits the grid's own
+    // padding inside the call, so the card clears both.
+    final gridPadding =
+        StreamCallParticipantsGridTheme.of(context).padding?.resolve(
+          Directionality.maybeOf(context),
+        ) ??
+        EdgeInsets.all(spacing.xs);
 
     return Align(
       alignment: Alignment.bottomCenter,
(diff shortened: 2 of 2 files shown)


renefloor:
@coderabbitai review

Brazol (approved):

PR #1349, opened 2026-09-15
fix(llc): ringing fixes for calls answered on the native call screen

Brazol:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability for calls answered through iOS CallKit during cold starts or while the app is terminated; calls are now detected and joined correctly.
  - Prevented duplicate call instances, repeated acceptance attempts, and duplicate call-screen navigation for the same incoming call.
  - Incoming-call state is now cleared when a call ends or is left.
  - Improved handling of calls accepted before joining, including failed acceptance cleanup and native call termination when acceptance cannot complete.
  - Reusing an incoming call now preserves the same call instance throughout the ringing flow.
  - Prevented older call events from clearing or replacing newer calls with the same call ID.


packages/stream_video/lib/src/stream_video.dart
```diff
@@ -1149,9 +1281,56 @@ class StreamVideo extends Disposable {
...
+    if (!isCallAcceptedOnThisDevice(cid)) return null;
+    return _findCall(cid);
+  }
+
+  /// The live [Call] for [cid], if this client already has one.
+  Call? _findCall(String cid) {
+    final ringing = _ringingCalls[cid] ?? _state.incomingCall.valueOrNull;
+    if (ringing?.callCid.value == cid) return ringing;
+    return activeCalls.firstWhereOrNull((call) => call.callCid.value == cid);
+  }
```
renefloor: **Critical — this can return the wrong `Call`, or `null`, and re-open the bug the PR fixes.**

`_locallyAcceptedCalls[cid]` *is* the accepted `Call` — `markCallAcceptedOnThisDevice` stored it. But this confirms the key exists and then goes looking for the call in three *other* registries. Two failure modes:

**Returns `null`** when the accepted `Call` is in none of them — e.g. a call the app made itself and accepted (`makeCall().accept()`), or the second instance in the two-instances-per-cid scenario your own test `'an older call tearing down leaves a newer one for the same cid alone'` constructs. Both callers (`_acceptIncomingCall` L1040, `consumeAndAcceptActiveCall` L832) then fall through to `consumeIncomingCall` + `call.accept()`; the second accept fails `Call.accept`'s status guard, and that failure branch calls `_endUnjoinableNativeCall` → **hangs up the live call**. That is exactly what `consumeAndAcceptActiveCall alongside the ringing accept flow` is written to prevent.

**Returns a *different* instance** when `_ringingCalls[cid]` holds a different `Call` than the accepted one. `onCallAccepted` then hands the app the unaccepted twin, whose `join()` is rejected as "a call with the same cid is in progress" — the failure mode your `'returns the same Call instance for repeated consumes'` comment describes.

The fix is also a simplification — it deletes `_findCall` entirely:

```suggestion
  /// The [Call] for [cid] when another entry point on this client already
  /// accepted it.
  Call? _acceptedElsewhereOnThisClient(String cid) => _locallyAcceptedCalls[cid];
```

Worth a regression test: accept a `Call` that is in neither `_ringingCalls` nor `state.incomingCall`, then assert `consumeAndAcceptActiveCall` hands back *that* instance and never calls `endCallByCid`.

---

**Why this happened, and the thing I'd actually change.** `_ringingCalls`, `_locallyAcceptedCalls` and `_acceptingCallCids` (L339-351) are three parallel cid-keyed registries with overlapping lifecycles, mutated from two classes — `Call` reaches back in via three `@internal` methods (`call.dart:954`, `2829-2830`). Their invariant, "the accepted call *is* the ringing call", is maintained by hand at every call site rather than structurally, and this function is where that hand-maintenance slipped.

A single `Map<String, _RingingFlow>` holding `call` / `acceptedHere` / `accepting` would make the invariant unstateable-wrong, collapse the three `@internal` mutators into one, and make `_acceptedElsewhereOnThisClient` a field read. Not a merge blocker on its own, but I'd rather see it now than after a fourth registry joins.

renefloor (commented):
## Review: ringing fixes for calls answered on the native call screen

The central insight — marking acceptance *before* the coordinator request, so the marker can never be observed too late — is correct, and the tests that pin it are genuinely strong. Extracting `isAcceptedOnThisDevice` as a pure function so the CallKit decision is testable without a platform channel is the right call, and `_endUnjoinableNativeCall` consolidates four scattered "the user is on an answered screen with nothing behind it" cases behind one named, documented call.

One blocker and three things I'd want resolved before merge, inline. The rest are suggestions.

**Blocker:** `_acceptedElsewhereOnThisClient` confirms the accepted call exists and then goes looking for it somewhere else, so it can return `null` or the wrong instance — which falls through to a second accept, and that failure path ends the live native call. That is verbatim the regression this PR fixes.

**Important:** a changelog entry describes a change that didn't happen; an already-answered call still raises `state.incomingCall`; `onCallAccepted` can now fire twice and that's a new, undocumented contract requirement for every integrator.

### Test coverage

Among the better test additions in this repo — the comments explain *why each case exists and what breaks without it* rather than restating the assertion. Gaps worth closing:

- The `null`-return case of `_acceptedElsewhereOnThisClient` (see inline). No test constructs an accepted call that is outside `_ringingCalls` / `incomingCall` / `activeCalls`.
- The new `_endUnjoinableNativeCall` branches in `consumeAndAcceptActiveCall`. Connect failure and consume failure now end the native call; on `main` a flaky cold-start connect just returned `false` and left CallKit up. That behaviour change is untested.
- The `_acceptingCallCids` concurrency guard — no test fires two accepts concurrently.
- The junction at `_verifyAndConsumeIncomingCall`, where a call is kept because `_isAnsweredOnThisDevice` is true via the new marker. The marker is tested and the accept path is tested, but not the point where they meet.

### CI

The red `stream_video_flutter` check is the pana score gate (110 < 120). It already fails on `main` (the run for #1323) and touches no file here — pre-existing, not a blocker for this PR.

### One question rather than a finding

The sample app now listens to `state.incomingCall` on **mobile** too (`app_content.dart:145`), where it previously didn't. A foreground ring on mobile will now push the call screen while CallKit / the notification is also showing it. Intentional in-app ringing UI, or a side effect of the refactor?

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -538,7 +556,17 @@ class StreamVideo extends Disposable {
         return;
       }
 
+      final consumedCall = _ringingCalls[event.data.callCid.value];
+      if (consumedCall != null) {
+        _logger.v(
+          () => '[onCoordinatorEvent] reusing consumed call: ${event.data}',
+        );
+        _state.incomingCall.value = consumedCall;
```
renefloor, line 581: **Important — an already-answered call still raises the app-facing incoming-call signal.**

This branch has no accepted-check. On the cold-start CallKit order (push → consume → accept → WS connects → `call.ring` arrives), it sets `incomingCall` to a call the user already answered and that is mid-`join()`. Any app rendering an in-app ringing UI from `state.incomingCall` shows a ring over an active call. Your test `'keeps the consumed call when the ringing event arrives after'` asserts exactly this publish.

This is why the sample app needed `_routedCallCid` at all — a guard here fixes it for every integrator instead:

```suggestion
      final consumedCall = _ringingCalls[event.data.callCid.value];
      if (consumedCall != null) {
        if (isCallAcceptedOnThisDevice(event.data.callCid.value)) {
          _logger.v(
            () => '[onCoordinatorEvent] already accepted here: ${event.data}',
          );
          return;
        }

        _logger.v(
          () => '[onCoordinatorEvent] reusing consumed call: ${event.data}',
        );
        _state.incomingCall.value = consumedCall;
        return;
      }
```

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -951,43 +1006,82 @@ class StreamVideo extends Disposable {
...
+    try {
+      _cancelIncomingAutoRejectTimerByCid(cid);
 
-    final acceptResult = await callToJoin.accept();
+      final accepted = _acceptedElsewhereOnThisClient(cid);
+      if (accepted != null) {
+        _logger.v(() => '[acceptIncomingCall] already accepted: $cid');
+        onCallAccepted?.call(accepted);
+        return;
+      }
```
renefloor, line 1069: **Important — `onCallAccepted` can now fire twice for one call, and that's a new undocumented contract.**

On an Android cold start, `consumeAndAcceptActiveCall` accepts and invokes `onCallAccepted`; the subsequent `ActionCallAccept` then reaches here, finds the call already accepted, and invokes `onCallAccepted` **again**.

Your own test comment on `'leaves an answered call alone on Android'` names this hazard — "Accepting here too would hand the app a second `onCallAccepted` and navigate it twice" — but that carve-out only covers the `verifyDisplayedIncomingCalls` path, not this one.

Net it's still a clear improvement over `main`, which hung the live call up instead. But it makes "your `onCallAccepted` must be idempotent" a requirement for every integrator, and the sample app absorbing it silently via `_routedCallCid` is the kind of thing integrators won't replicate on their own. Worth stating in the dartdoc on `observeCoreRingingEvents` and in the changelog.

Separately: this early return calls `onCallAccepted` but skips the `unawaited(callToJoin.join())` the full path performs. If the first accepter was `consumeAndAcceptActiveCall` — which never joins — nothing in the SDK joins the call.

packages/stream_video/CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 ### 🐞 Fixed
 
+- Fixed `StreamVideo.state.incomingCall` still holding a call after it ended.
```
renefloor, line 15: **Important — this entry describes a change that didn't happen.**

> `consumeIncomingCall` no longer publishes to `StreamVideo.state.incomingCall`.

It never did. On `main` it only *read* `incomingCall` (`stream_video.dart:1423` on the base), and `_makeCallFromRinging` doesn't publish either. Meanwhile the incoming call still *is* published for a consumed call — by the ringing event path at `stream_video.dart:559-564`, which `'keeps the consumed call when the ringing event arrives after'` asserts.

An integrator reading this will expect their in-app incoming UI to stop firing on the CallKit cold-start path. It won't. What actually changed is instance identity:

```suggestion
- Fixed `StreamVideo.state.incomingCall` still holding a call after it ended.
- Fixed an issue where `consumeIncomingCall` could create multiple `Call` instances for the same ringing flow, causing state conflicts and UI issues.
- `consumeIncomingCall` and the `call.ring` event now share one `Call` per ringing flow, so `StreamVideo.state.incomingCall` no longer switches to a second instance for a call already consumed from a push.
```

(That also drops the trailing whitespace on the first line.)

Also: the `consumeIncomingCall` entry was copied into `packages/stream_video_push_notification/CHANGELOG.md:9`, but that fix lives entirely in `stream_video`.

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -791,54 +819,75 @@ class StreamVideo extends Disposable {
...
-    final acceptResult = await call.accept();
-    if (acceptResult.isFailure) {
-      _logger.w(
-        () =>
-            '[consumeAndAcceptActiveCall] error accepting call: '
-            '${acceptResult.getErrorOrNull()}',
-      );
+      final call = callResult.getDataOrNull();
+      if (call == null) {
+        await _endUnjoinableNativeCall(cid);
```
renefloor, line 892: Suggestion: ending the user's native call with no log line. Every other `_endUnjoinableNativeCall` site logs why first — this one and the matching branch at L1063-1064 don't, so a hung-up CallKit call leaves nothing behind to explain it.

Both branches are also effectively dead: `isFailure` was checked immediately above, so `getDataOrNull()` can't be null here. Either drop them or log before ending.

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -1000,18 +1094,18 @@ class StreamVideo extends Disposable {
...
-  /// are therefore verified directly, so a ringing flow that was already
-  /// resolved is dismissed instead of ringing until the native timeout.
+  /// On iOS, CallKit may show the incoming call UI before Dart code runs,
+  /// so ringing events can be missed. This inspects all displayed calls:
+  /// - Still-ringing calls are verified so dismissed flows are ended.
+  /// - Calls already answered but not picked up by the app are accepted and joined.
   ///
   /// Called automatically by [observeCoreRingingEvents].
-  Future<void> verifyDisplayedIncomingCalls() async {
+  Future<void> verifyDisplayedIncomingCalls({
```
renefloor, line 1133: Suggestion: `verifyDisplayedIncomingCalls` no longer matches its name or its documented contract — it now accepts and joins calls.

It's public, and the doc only says "Called automatically by `observeCoreRingingEvents`". Called manually without `onCallAccepted`, it will accept and `join()` a call the app is never told about.

The doc bullet "Calls already answered but not picked up by the app are accepted and joined" is also unconditional, while `_acceptDisplayedIncomingCall` bails on every non-iOS platform (L1145) — and that bail is the only early return in the function without a log, so on Android it's invisible. Worth stating the iOS-only restriction in the doc and logging the bail.

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -490,7 +502,9 @@ class StreamVideo extends Disposable {
       await _client.disconnectUser();
       _subscriptions.cancelAll();
 
-      // Resetting the state.
+      _ringingCalls.clear();
+      _locallyAcceptedCalls.clear();
+      _acceptingCallCids.clear();
```
renefloor: Suggestion: `dispose()` doesn't clear these three. It clears `_incomingAutoRejectTimers` and `_state` but not `_ringingCalls` / `_locallyAcceptedCalls` / `_acceptingCallCids`. Same for `_disconnect()`'s own `isDisconnected` early return above, which returns success without reaching here.

The instance is being discarded either way so it isn't a live leak, but these hold strong `Call` references and the asymmetry with the timers will bite whoever touches this next.

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -951,43 +1006,82 @@ class StreamVideo extends Disposable {
...
+  }) async {
+    if (!_acceptingCallCids.add(cid)) {
+      _logger.v(() => '[acceptIncomingCall] already accepting: $cid');
       return;
     }
 
-    final callToJoin = consumeResult.getDataOrNull();
-    if (callToJoin == null) return;
+    try {
+      _cancelIncomingAutoRejectTimerByCid(cid);
```
renefloor: Suggestion: this moved behind the `_acceptingCallCids` dedupe guard, so a concurrent accept now returns early without cancelling the timer — and `consumeAndAcceptActiveCall` never cancels it at all.

Harmless today only because the timer body re-checks `status.acceptedByMe` before rejecting. That's an accident rather than a design; a comment saying the timer is self-guarding, or cancelling before the guard, would keep it that way.

packages/stream_video/lib/src/call/call.dart
```diff
@@ -2824,11 +2826,17 @@ class Call {
...
     await clearE2EEManager();
 
+    _streamVideo.clearCallAcceptedOnThisDevice(callCid, this);
+    _streamVideo.releaseRingingCall(callCid, this);
     await _streamVideo.state.removeActiveCall(this);
     if (_streamVideo.state.outgoingCall.valueOrNull?.callCid == callCid) {
       await _streamVideo.state.setOutgoingCall(null);
     }
 
+    if (identical(_streamVideo.state.incomingCall.valueOrNull, this)) {
```
renefloor, line 2836: Nit: `identical(...)` here, but `_resolveStaleIncomingCall` (`stream_video.dart:1365`) does the same check with `==`. Equivalent, since `Call` doesn't override `==` — worth picking one so the next reader doesn't go looking for the distinction.

renefloor (commented):
## Re-review @ `0302902c`

Checked out the head and ran both new suites locally: **21/21 pass**. I also confirmed the new `'hands back an accepted call held in no other registry'` test is a real regression test — it fails against the old `_findCall` rather than passing vacuously.

### Resolved (7 of 9)

The blocker is properly fixed: `_acceptedElsewhereOnThisClient` is now `=> _locallyAcceptedCalls[cid]` and `_findCall` is gone. Three others landed **better than what I proposed**:

- **The `_onEvent` guard** went *above* the `_ringingCalls` lookup rather than inside it, so it also covers the `_makeCallFromRinging` fallback — a case my suggestion would have missed.
- **The auto-reject cancel** moved ahead of the dedupe guard in `_acceptIncomingCall` *and* was added to `consumeAndAcceptActiveCall` before the slow `connect()`, with a comment in the timer body documenting the self-guard. That closes the hole instead of just annotating it.
- **`_clearRingingState()`** was extracted and applied to `dispose()`, `_disconnect()`'s early return, *and* `_handledIncomingCallCids`, which I hadn't flagged.

### Two threads were resolved but not addressed

I've re-opened them with a note — both are documentation rather than code, so I couldn't tell from the diff whether they were declined or missed:

- `CHANGELOG.md:15` still claims `consumeIncomingCall` no longer publishes to `state.incomingCall`.
- `observeCoreRingingEvents` still has no note that `onCallAccepted` may fire twice, and the early return still skips the `join()` the full path performs.

Happy for either to be closed as "won't do" — just want it to be a decision rather than an oversight.

### One new finding

A check-then-act race across the `await` in `consumeIncomingCall`, inline below. Same failure mode this PR exists to fix — narrowed, not eliminated. CodeRabbit spotted this one too; I verified it independently.

### Still open from last time

Three test gaps (the `_endUnjoinableNativeCall` branches in `consumeAndAcceptActiveCall`, the `_acceptingCallCids` concurrency guard, the `_isAnsweredOnThisDevice`-via-marker junction), and my question about the sample app now listening to `state.incomingCall` on mobile.

The three-registry consolidation I bundled into the critical comment wasn't done, but `_clearRingingState()` centralises the lifecycle and the bug it caused is gone — I'd leave that alone.

Red `stream_video_flutter` is still the pana gate that already fails on `main`. Unrelated.

packages/stream_video/lib/src/stream_video.dart
```diff
@@ -1449,6 +1653,8 @@ class StreamVideo extends Disposable {
       preferences: preferences ?? _options.defaultCallPreferences,
     );
```
renefloor, line 1665: **New finding — check-then-act race across the `await`, same failure mode this PR fixes.**

The cache is read at L1626, before `await _client.getCall(...)`, and written here at L1655 after it. Two concurrent calls for the same cid that both miss the read will both await the network, both build a `Call`, and the second write clobbers the first — leaving the first caller holding an orphan. That is precisely the *"multiple `Call` instances for the same ringing flow"* failure the changelog entry above describes: narrowed by the caching, not eliminated.

It's reachable because the two guards don't compose — `_handleIncomingCall` uses `_handledIncomingCallCids`, `_acceptIncomingCall` uses `_acceptingCallCids`, so an `ActionCallIncoming` and an `ActionCallAccept` for the same cid can both be inside `consumeIncomingCall` before either `getCall` returns. On an iOS cold start that's the ordinary sequence: the incoming push starts a consume, the user answers a moment later and starts another. `consumeIncomingCall` is also public, so an app can race the SDK's own path.

Minimal fix — re-check before building:

```suggestion
    // Re-check: a concurrent consume for the same cid may have populated the
    // cache while the metadata fetch above was in flight.
    final cached = _ringingCalls[cid];
    if (cached != null) return Result.success(cached);

    _ringingCalls[cid] = call;
```

That leaves a throwaway `Call` built on the losing path. If you'd rather not build one at all, memoise the in-flight future per cid (`Map<String, Future<Result<Call>>>`, cleared in a `whenComplete`) and have every caller await the same one — which also makes `_acceptingCallCids` and `_handledIncomingCallCids` redundant for this particular hazard.

Either way it deserves a test: two `consumeIncomingCall` futures for one cid started before the `getCall` completer resolves, asserting both resolve to the same instance.

packages/stream_video/CHANGELOG.md
renefloor: Still open at `0302902c` — the entry is unchanged, so flagging in case this was resolved by the batch rather than declined.

The claim is that `consumeIncomingCall` *stopped* publishing to `state.incomingCall`. It never published: on `main` it only read that field, and `_makeCallFromRinging` doesn't publish either. What this PR actually changed is (a) instance identity — one `Call` per ringing flow — and (b) the new accept guard in `_onEvent`, which is the thing that genuinely stops a second incoming UI.

So the sentence describes a change that didn't happen while the real one goes unmentioned. Suggested replacement:

```markdown
- `consumeIncomingCall` and the `call.ring` event now share one `Call` per ringing flow, so `StreamVideo.state.incomingCall` no longer switches to a second instance for a call already consumed from a push.
- A `call.ring` event for a call already answered on this device no longer raises `StreamVideo.state.incomingCall`, so an app rendering its own incoming UI from it does not show a ring over a call that is already being joined.
```

Also still duplicated into `packages/stream_video_push_notification/CHANGELOG.md:9`, where the fix doesn't live.

Fine to close as "won't do" — just want it to be a decision.

packages/stream_video/lib/src/stream_video.dart
renefloor: Still open at `0302902c` — `observeCoreRingingEvents` is unchanged (still *"Helper method to observe core ringing events."*), so flagging in case this was resolved by the batch.

Both halves stand:

1. **Double `onCallAccepted`.** On an Android cold start `consumeAndAcceptActiveCall` accepts and invokes the callback; the subsequent `ActionCallAccept` reaches `_acceptIncomingCall`, finds the call already accepted, and invokes it again. The sample app absorbs this with `_routedCallCid`, but integrators won't know to — it needs to be on the dartdoc.

2. **The early return skips `join()`.** The full path ends with `unawaited(callToJoin.join())`; the already-accepted branch doesn't. If the first accepter was `consumeAndAcceptActiveCall`, which never joins either, nothing in the SDK joins the call — it only works today because the integrator's call screen joins on mount.

(2) in particular is worth a second look, since it's behavioural rather than documentation.

renefloor (approved):

PR #1354, opened 2026-09-16
fix(llc,ui): stop re-emitting participant state for events that change nothing

renefloor:
### 🎯 Goal

Livestreams with many joining participants produce a high rate of SFU events (`ParticipantJoined`, `AudioLevelChanged`, `ConnectionQualityChanged`, `ParticipantUpdated`). The Flutter SDK had no throttling or batching anywhere in that path, and every handler rebuilt the whole participant list and pushed a new `CallState`, so listeners woke on every audio level tick whether or not anything they render had changed.

stream-video-swift and stream-video-android both solve this the same way — adaptive, count-based throttling of participant updates — and neither does anything special on the WebRTC side. This ports that approach.

Closes FLU-793.

### 🛠 Implementation details

**Change guards in the SFU state handlers** (`state_sfu_mixin.dart`)

`sfuUpdateAudioLevelChanged`, `sfuConnectionQualityChanged`, `sfuDominantSpeakerChanged`, `sfuInboundStateNotification` and `sfuParticipantUpdated` now keep the existing `CallParticipantState` instances and skip the state write entirely when the event leaves everything unchanged. The audio level rule is Swift's: a silent participant who was already silent is skipped.

The skip is not a flag each handler carries. They hand a mapping function to `_updateParticipants`, which returns the participant untouched to mean "nothing to do" and decides from identity — the same thing the sorting mixin and `partialState` already read to detect a change, so there is nothing to keep in sync and nothing to forget to set. It also builds no list when nothing changed, where each handler previously allocated one per event and discarded it; that path is every audio level tick.

Matching goes through a map built once per event, replacing a `firstWhereOrNull` over the event payload per participant (`sfuUpdateAudioLevelChanged` was O(n·m)).

Preserving instance identity is what makes the rest of the PR work — downstream `distinct` and the widget-level checks can then use identity rather than deep equality over 22 `Equatable` props per participant.

**`Call.participantsStream`** (new)

A participant list stream whose emission interval grows with the participant count, with the tiers taken from Swift's `CollectionDelayedUpdateObserver`:

| participants | interval |
| --- | --- |
| < 16 | 16 ms |
| < 50 | 250 ms |
| < 100 | 500 ms |
| >= 100 | 1 s |

Trailing only: each window emits the most recent list to arrive during it, so a change is never dropped, only collapsed with the ones around it.

The window is a hand-written `StreamTransformer` rather than rxdart's `throttle`, after three separate problems with that operator on this path:

- **Leading emissions doubled the rate.** `eventAfterLastWindow` closes a window before the next value reopens it, so a continuous source gets the trailing value *and* the next value as the new window's leading one. Measured at 9 emissions per 500 ms against a 100 ms window, which made every tier mean half what it says.
- **It never completed when the source closed with no window running** — the normal case, since a stable participant list means an idle throttle. Any consumer using `await for`, `.last` or `.drain()` would hang with no error.
- **It dropped a lone held value on close**, gated on `queue.length > 1`. The disconnect that clears `callParticipants` could land inside a window and never reach a listener.

The last two came out of review and are covered by tests now. Combine's `throttle(latest:)` emits once per interval, and this matches.

The interval is evaluated once per window, when it opens; a value arriving mid-window does not re-measure it, so a change in participant count takes effect on the next window. Same as Swift, which recomputes its interval after each emission rather than mid-flight.

Where this lives matters, and both natives agree: Swift wires `CollectionDelayedUpdateObserver` up in `CallController` and Android keeps `TaskSchedulerWithDebounce` in core `CallState` — neither UI module throttles anything. So the policy sits on `Call`, `StreamCallParticipants` and the Android PiP overlay just subscribe, and `throttleByCollectionSize` is an unexported implementation detail, matching `CollectionDelayedUpdateObserver` not being public in Swift.

One window is shared by every listener, behind a `BehaviorSubject` held as a `late final` field. Sharing is the point: Swift feeds one `CollectionDelayedUpdateObserver` into `Call.state.participantsMap` and Android has one `StateFlow`, so every consumer sees the same list in the same frame. A per-listener variant was tried and reverted — it let two widgets rendering one call drift apart, and re-ran the upstream map and `ListEquality` distinct once per listener instead of once. The subject is not handed to listeners directly. It carries the list the last window closed on, which is older than the state a listener is starting from, so subscribing to it raw walked the list backwards for up to an interval — a just-joined participant appeared for one frame and then dropped off screen. `participantsStream` is a `late final` broadcast stream instead: every listener is given the live `CallState.callParticipants` first, the subject's replay is dropped, and the shared window drives everything after that. Being a field rather than a fresh stream per access is what keeps `StreamBuilder` from resubscribing on every rebuild.

`CallState.callParticipants` stays immediate, and deliberately so. Five call sites read it synchronously — `_getTrackForParticipant` starts a track straight off a `TrackPublished` event, the ringing flow checks who is left, and dynascale, the rtc manager and the participant mapper all do lookups. Throttling it would make a just-joined participant's track fail to start. Swift has the same split: `WebRTCStateAdapter.participants` is immediate and only its projection into `Call.state.participantsMap` is throttled.

**`CallPreferences.participantsThrottleIntervalResolver`** (new)

Takes the participant count and returns a duration, which is the shape the default tiers already had — so an integrator can shift the thresholds rather than being forced to a single constant. `defaultParticipantsThrottleInterval` is public and is the literal default, so it can be wrapped (`(n) => defaultParticipantsThrottleInterval(n) * 2`). Pass `null` to emit every change. Neither native SDK exposes this.

It is named for what it is — a function, like the existing `encryptionKeyResolver` — rather than for the `Duration` the other `*Interval` preferences hold. Adding it to the `CallPreferences` interface is source-breaking for an app with its own implementation, so it is under `⚠️ Breaking` in the changelog.

```dart
DefaultCallPreferences(
  participantsThrottleIntervalResolver: (_) => const Duration(milliseconds: 100),
)
```

The preference is read once, the first time `participantsStream` is accessed — the same way `callStatsReportingInterval` is read at session start. A later `updateCallPreferences` does not change it for that call. A resolver that throws surfaces as an error on the stream rather than leaving it stalled with no window armed, and a negative duration is treated as zero.

**Participant sorting** (`call_participants_sorting_mixin.dart`)

`_sortedParticipantKeys.indexOf(...)` was called inside the sort comparator, allocating a `'$userId-$sessionId'` string per call, on top of an O(n²) `indexOf` loop — O(n² log n) overall. Now a precomputed order map, so O(n log n). `setState` is skipped when the resulting list is unchanged, mirroring Android's `SortedParticipantsState.resort()`.

**The livestream widgets** (`livestream_content.dart`, `livestream_backstage_content.dart`)

Both selected a whole participant list inside a record. Dart compares a record's `List` field by identity, so neither hit the `ListEquality` path in `partialCallStateStream`'s distinct — they rebuilt on every participant update. The two need opposite fixes, which is worth spelling out because it's the general rule:

- `LivestreamContent` renders per-participant widgets, so it genuinely needs the list. Split: `status` still comes off the raw state, so a disconnect is acted on at once, and the participants come through `Call.participantsStream` via a new `CallParticipantsBuilder`. Only the reindent moved; no logic changed.
- `LivestreamBackstageContent` renders `participants.length` and nothing else, so it now selects the count. An `int` in a record compares by value, which makes the distinct work and keeps the number prompt. Throttling it would have been *worse* — a delayed count for no benefit.

`CallParticipantsBuilder` seeds `initialData` from the current call state, so a consumer doesn't wait out the first throttle window — at the 1 s tier that would leave a livestream blank. Its doc points people at `PartialCallStateBuilder` when they only need a derived value, which is the mistake backstage made.

The other three `partialState` consumers were checked and left alone: `call_content.dart:160` and `stream_picture_in_picture_android_view.dart:85` select only scalars, so their records compare by value; `call_content.dart:281` selects `state.localParticipant`, a single participant compared via Equatable — and a local mute needs to be prompt, not throttled.

**Bugs found along the way** (`call_participant_state.dart`, `state_sfu_mixin.dart`)

- `copyWithUpdatedAudioLevels` did `final levels = audioLevels; levels.add(...)` — mutating the list in place, so the "previous" immutable snapshot's history changed underneath it. Now copies, and `audioLevels` is copied into an unmodifiable list so neither the next writer nor a caller still holding the list it passed in can write through it. Instance identity is what every guard in this PR rests on.
- `sfuPinsUpdated` re-pinned every already-pinned participant with a fresh `DateTime.now()`. `pinnedAt` is what orders pinned participants in the `pinned` comparator, so a pins event reshuffled them. Now it keeps the existing server pin.
- `sfuDominantSpeakerChanged`'s first guard used `firstWhereOrNull`, which assumes one flagged participant — nothing enforces that, and `sfuJoinResponse` and `sfuParticipantUpdated` both take the flag off the wire. With two flagged and the event matching the first, a stale flag survived. It now compares each participant's desired flag against the one they carry, which handles several flagged participants without treating that as a case of its own.
- `CallParticipantsBuilder` read `participantsStream` in `build` while it was briefly a getter, so `StreamBuilder` resubscribed on every rebuild and restarted the throttle window. An ancestor rebuilding faster than the interval starved the list. It holds its stream in state now.

**Public behaviour change:** a silent participant's `audioLevel` and `audioLevels` now hold at the reading that took them below the speaking threshold, instead of tracking every quiet sample after it. That is what makes the audio-level guard worth having, and it is what Swift does. It is in the changelog.

**From review** (`call.dart`, `call_participant_state.dart`, `adaptive_throttle.dart`, `partial_call_state_builder.dart`, `call_participants.dart`)

- The stale-replay bug above, which every consumer of `participantsStream` hit.
- `_sealLevels` wrapped the caller's list in an `UnmodifiableListView` rather than copying it. A view writes through, so `CallParticipantState(audioLevels: myList)` left the caller able to mutate the "sealed" field with identity unchanged — exactly what the seal exists to prevent.
- The interval resolver was called inside the source's `onData`, so a throw went to the zone rather than the stream: no window armed, no value and no error downstream, and the participant list stopped updating for the rest of the call with a clean log.
- `PartialCallStateBuilder` cast `snapshot.data as T` on an error snapshot, turning a real stream error into a `TypeError` naming a cast. It now falls back to the current state and logs. `CallParticipantsBuilder` already fell back but discarded `snapshot.error`; it logs it now.
- `StreamCallParticipants.didUpdateWidget` cancelled its subscription when a controlled `participants` list was supplied and never re-took it when the widget went back to the call's own list, freezing the list silently. Pre-existing; fixed here because the same branch moved.
- Docs corrected where they described something other than the code: the preference's "read each time", the audio-level comment naming the wrong reading, the `audioLevel`/`audioLevels` field docs (neither "latest" nor plainly "the last 10" any more, and now unmodifiable), the `// ignore: close_sinks` justification, and the sorting mixin's identity claim.

### 🎨 UI Changes

No intended visual change. Participant list updates land up to one interval later in large calls, by design.

### 🧪 Testing

`melos run analyze` is clean on both packages. `stream_video`: 603 tests pass, including 46 new ones:

- `test/src/call/state/state_sfu_mixin_test.dart` — each guarded handler emits on a real change and keeps the list identical on a no-op; untouched participants keep their instance; the audio level history is not shared with the previous snapshot and cannot be mutated; `sfuInboundStateNotification` multi-track grouping and the pause→unpause round trip; pin stability; and the two-flagged-dominant-speaker case. Plus the cases a guard could be "simplified" into: a speaker falling silent is still written through, a local pin survives a server pins event, an `unspecified` quality does not downgrade a known one, and a no-op event pushes no new `CallState` — that last one asserted on the state stream, since the setter writes unconditionally and an identical list does not prove nothing was re-emitted.
- `test/src/utils/adaptive_throttle_test.dart` — rebuilt on `fakeAsync`, so the timing assertions are exact rather than wall-clock. Covers the tier boundaries (15/16, 49/50, 99/100), once-per-window under a continuous source, window measurement from the opening list, error forwarding, cancel teardown, the three completion cases, and an interval callback that throws or returns a negative duration. Completion runs on the real event loop, since it is about ordering rather than timing.
- `test/src/call/call_participants_stream_test.dart` — stream identity across accesses, replay to a late listener, and two listeners receiving identical values. Now driven by real state changes rather than an empty call, so it covers the window collapsing several updates into one, `null` bypassing the throttle, and a listener subscribing mid-window starting from the live list. That last one fails on the pre-fix stream with `[0]` against two live participants.

`stream_video_flutter` had no participant tests before this PR. It now has the sorting mixin (append-on-join, order preserved on leave, sort stability, comparator precedence, filtering, skip-`setState`) and `CallParticipantsBuilder` (one stream across rebuilds — which fails on the pre-fix widget with 6 accesses instead of 1 — delivery while an ancestor rebuilds, re-taking the stream when the call changes, and falling back to the current state when the stream errors). 16 tests pass.

The `update_goldens` workflow was run on `8bd7dc8b` ([run 35114738369](https://github.com/GetStream/stream-video-flutter/actions/runs/35114738369)) and regenerated every CI golden byte-identically — "Working tree clean. Nothing to commit." The commits since then are state, docs and tests only, and a local `--update-goldens` on the current head leaves every committed PNG unchanged, so nothing in this PR changes what gets rendered.

A fresh checkout has no `goldens/macos` directory — only `goldens/ci` is committed — so the macOS variants of `call_content_test.dart` fail until `flutter test --tags golden --update-goldens` has run once locally. The CI variants pass either way.

**No benchmark was run**, so this PR claims no measured improvement over `main`. The changes are structural; someone should measure a real livestream before we describe this as a speedup anywhere user-facing.

### 📋 Deliberately out of scope

Two items from the FLU-793 analysis are not here, both because they want the `Map<String, CallParticipantState>` rework planned for `v2` first:

- **Android-style batching of `ParticipantJoined` at ingestion.** `call_session.dart:770` runs `stateManager.sfuParticipantJoined(event)` and then `await rtcManager?.flushPendingDecryptors()` under the same lock. Delaying ingestion would leave a join's decryptor flush with no participant to attach to — harmless with E2EE off, where it early-returns, but not with it on. Doing it properly means moving the flush onto a "batch applied" signal. A join storm therefore stays O(n²) in the state layer.
- **Unwinding `_sfuEventsLock`.** It also guards peer-connection ordering; low payoff for the risk in this PR.
- **`PartialCallStateBuilder` resubscribing on every rebuild.** `partialState` builds a fresh chain per call and it is read in `build`. Holding it in `State` was tried and reverted: Dart does not canonicalise the inline closures every call site passes as `selector`, so the widget re-takes the stream on each rebuild regardless, and keying off the selector only adds a way to ignore one that genuinely changed. Fixing it properly means the call sites hoisting their selectors.

#### General

### ☑️Reviewer Checklist
- [ ] Sample runs & works

## Summary by CodeRabbit

- **New Features**
  - Added `Call.participantsStream` with configurable participant-count-based throttling or immediate updates.
  - Added Flutter support for rendering participants from the throttled stream while preserving current-state responsiveness.
  - Preserved participant ordering and reduced unnecessary widget rebuilds.

- **Bug Fixes**
  - Made audio history read-only and preserved the last audio level while participants are silent.
  - Prevented redundant state updates and corrected server-pinned participant ordering.
  - Improved widget error fallback and rendering stability.

- **Breaking Changes**
  - Custom `CallPreferences` implementations must provide the participant throttling resolver.


packages/stream_video/lib/src/call/call.dart
```diff
@@ -432,6 +434,98 @@ class Call {
...
+        var latest = _stateManager.callState.callParticipants;
+        controller.add(latest);
+
+        var replayed = false;
+        final subscription = _participantsSubject.stream.listen(
+          (value) {
+            // The subject always opens with its current value.
+            if (!replayed) {
+              replayed = true;
+              return;
```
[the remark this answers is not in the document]
renefloor: Correct, and fixed. `BehaviorSubject` caches its latest error as well as its latest value, so a listener can open on `onError` rather than `onData` — which left `replayed` false and made the next real participant list get eaten as if it were the replay. `onError` now sets the flag before forwarding.

Verified rather than assumed: the regression test `delivers the next value after an error is replayed` fails against the previous code with `Expected: [1] / Actual: []`.

packages/stream_video/lib/src/models/call_participant_state.dart
```diff
@@ -81,10 +83,19 @@ class CallParticipantState extends Equatable
   final SfuParticipantSource? participantSource;
   final bool isOnline;
 
-  /// The latest audio level for the user.
+  /// The user's most recent audio level while they were above the speaking
+  /// threshold.
+  ///
+  /// Updates stop while a participant is silent, so this holds at the reading
+  /// that took them below the threshold rather than tracking every quiet
+  /// sample. Use [isSpeaking] to tell the two apart.
```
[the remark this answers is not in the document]
renefloor: Agreed — the first sentence contradicted the paragraph under it. It now reads "The most recent audio level retained for the user.", with the threshold detail left to the paragraph below.

packages/stream_video/lib/src/models/call_participant_state.dart
```diff
@@ -103,6 +114,17 @@ class CallParticipantState extends Equatable
   final ViewportVisibility screenShareViewportVisibility;
 
   bool get isPinned => pin != null;
+
+  /// Identity is used all over the SDK to tell whether a participant changed,
+  /// which only holds while nothing mutates a collection in place. Handing out
+  /// an unmodifiable view makes that an error rather than a silently stale UI.
+  static List<double> _sealLevels(List<double> levels) {
+    if (levels is UnmodifiableListView<double>) return levels;
```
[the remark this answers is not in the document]
renefloor: Right, and thanks for catching it — the copy was added but the shortcut left in, which defeats it. An `UnmodifiableListView` a caller builds over their own list still writes through, so trusting the type told us nothing. `_sealLevels` copies unconditionally now.

Took the "remove the shortcut" option rather than distinguishing internally sealed lists: `audioLevels` is capped at 10 entries, so the copy is negligible and not worth a private subclass to skip. Covered by a test that mutates the backing list behind a view it was handed.

packages/stream_video_flutter/lib/src/widgets/partial_call_state_builder.dart
```diff
@@ -27,6 +27,66 @@ class PartialCallStateBuilder<T> extends StatelessWidget {
...
+    if (widget.call != oldWidget.call) {
+      _participants = widget.call.participantsStream;
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return StreamBuilder<List<CallParticipantState>>(
+      stream: _participants,
+      initialData: widget.call.state.value.callParticipants,
```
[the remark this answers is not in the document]
renefloor: Confirmed and fixed with the key you suggested. `StreamBuilder` carries its snapshot through `afterDisconnected` on a stream swap and only reads `initialData` in `initState`, so the previous call's participants really did render for a frame.

Your note that the existing changed-call test only checks stream access was also right. The new test `does not show the previous call on the frame the call changes` fails without the key with `Expected: ['bob'] / Actual: ['alice']`.

packages/stream_video/lib/src/models/call_preferences.dart
```diff
@@ -62,6 +63,20 @@ abstract class CallPreferences {
...
+  ///
+  /// A large call emits participant updates far faster than a screen can
+  /// usefully repaint, so the default returns a longer interval the more
+  /// participants there are. Return a shorter one to trade CPU for latency, or
+  /// set this to `null` to emit every update.
+  ///
+  /// This does not affect `CallState.callParticipants`, which is always
+  /// up to date. It is read each time `Call.participantsStream` is used, so
+  /// changing it through `updateCallPreferences` reaches new listeners but
+  /// leaves existing ones on the interval they subscribed with.
```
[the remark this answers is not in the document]
renefloor: Already addressed. The doc now says the resolver is read once, the first time `Call.participantsStream` is accessed, and that a later `updateCallPreferences` has no effect for that call.

Went with documenting the first-access behaviour rather than rebuilding the shared stream: rebuilding would swap the window out from under existing listeners, and one shared window is the property the rest of the design leans on. (The preference is also renamed to `participantsThrottleIntervalResolver`, since it is a function rather than the `Duration` the sibling `*Interval` preferences hold.)

renefloor:
Re: CodeRabbit's outside-diff finding, *Rebind the PiP participant subscription when `call` changes* (`android_pip_overlay.dart:58-60`).

**The finding is accurate, and I'm skipping it here.** Both halves check out against the current code:

- `_AndroidPipOverlayState` is `initState` → `dispose` → `build`, with no `didUpdateWidget`, so a changed `Call` leaves `_participantsSubscription` on the old one while `build` reads `widget.call` for the tile.
- `_StreamPictureInPictureAndroidViewState.didUpdateWidget` does handle `widget.call != oldWidget.call` for `_callStateSubscription`, but never marks `_overlayEntry` for rebuild. The entry passes `call: call`, a getter reading `widget.call`, so this isn't a captured stale reference — it's an entry that has no reason to re-run its builder. "Mark for rebuild" is the right description.

**But it isn't this PR's.** All this PR changes on those lines is the stream source inside the existing `initState`:

```dart
-    _participantsSubscription = widget.call
-        .partialState((state) => state.callParticipants)
-        .listen(recalculateParticipants);
+    _participantsSubscription = widget.call.participantsStream.listen(
+      recalculateParticipants,
+    );
```

The missing `didUpdateWidget` and the un-refreshed overlay entry are identical before and after — which is presumably why the bot itself filed this as outside the diff range. Reachability is also narrow: it needs the same element rebuilt with a different `Call`, which I don't think the SDK's own widgets do. The best argument for fixing it is the inconsistency — that `didUpdateWidget` already treats a changed call as real for one subscription and ignores it for the others.

Worth its own change rather than riding along on a throttling PR. (Correcting myself: I first said there was no fixture for it — there is, `android_pip_overlay_test.dart`.)

**Both halves are now fixed in #1357, on `v2`.**

- The same gap existed in the iOS `StreamPictureInPictureUiKitView`, and #1357 was making it sharper — it binds a `ViewportHandle` to `widget.call` on first build, so a changed call would keep it reporting into a call it had left. Fixed in `8f0475a1`.
- The Android overlay and the `markNeedsBuild()` on the entry: `db0415fe`, with a test that swaps the call under the overlay and expects the participants to change with it.

It went to `v2` rather than `main` because that is where the work is focused, and `main` is being kept to minimal changes. So nothing is needed here — this PR's diff on those lines stays as it is.

packages/stream_video_flutter/lib/src/call_participants/call_participants.dart
```diff
@@ -127,9 +127,9 @@ class _StreamCallParticipantsState extends State<StreamCallParticipants>
     );
 
     if (widget.participants == null) {
-      _participantsSubscription = widget.call
-          .partialState((state) => state.callParticipants)
-          .listen(recalculateParticipants);
+      _participantsSubscription = widget.call.participantsStream.listen(
```
Brazol: **`participantsStream` errors are unhandled here — this is a crash path.**

`adaptive_throttle.dart` deliberately catches a throwing `participantsThrottleIntervalResolver` and calls `controller.addError`, with the comment that it would otherwise "reach the zone and leave this stream silently stalled". `Call._buildParticipantsStream` then forwards that error to every listener.

This `.listen` (and the one at line 161, and `android_pip_overlay.dart:58`) passes no `onError`, so the error goes straight back to the zone as an uncaught async error — an app-level crash, on the one path the PR added error plumbing for.

`CallParticipantsBuilder` and `PartialCallStateBuilder` both handle `snapshot.hasError`. These three subscriptions should too — log it and keep the last known list.

packages/stream_video_flutter/lib/src/call_participants/call_participants.dart
```diff
@@ -145,18 +145,22 @@ class _StreamCallParticipantsState extends State<StreamCallParticipants>
...
+      _participantsSubscription = null;
 
       if (!const ListEquality<CallParticipantState>().equals(
         widget.participants!.toList(),
         oldWidget.participants?.toList(),
       )) {
         recalculateParticipants(widget.participants!);
       }
-    } else if (widget.call != oldWidget.call) {
+    } else if (widget.call != oldWidget.call ||
```
Brazol, line 179: **A `sort` / `filter` change can now go unapplied indefinitely.**

`didUpdateWidget` re-runs `recalculateParticipants` only when `widget.participants` or `widget.call` changes — never when `widget.sort` or `widget.filter` changes.

That was survivable before this PR: `sfuUpdateAudioLevelChanged` wrote state every ~100 ms, so the next `recalculateParticipants` picked up the new comparator almost immediately. This PR stops those writes when nobody is speaking — which is the point of it — so in a quiet call an app that swaps its sort comparator now sees nothing happen until someone talks or joins.

The fix is one clause in the `else if`. One thing to weigh: if an integrator passes a closure built inside `build()`, `!=` is true on every rebuild and you would re-sort each time. The new identity guard suppresses the `setState`, so it is cheap-ish, but not free.

packages/stream_video/lib/src/models/call_participant_state.dart
```diff
@@ -103,6 +113,16 @@ class CallParticipantState extends Equatable
   final ViewportVisibility screenShareViewportVisibility;
 
   bool get isPinned => pin != null;
+
+  /// Identity is used all over the SDK to tell whether a participant changed,
+  /// which only holds while nothing mutates a collection in place. Handing out
+  /// an unmodifiable view makes that an error rather than a silently stale UI.
+  static List<double> _sealLevels(List<double> levels) {
```
Brazol, line 120: **This copies the audio history on every `copyWith`, including the ones that never touch audio.**

`copyWith`, `copyWithPin` and `copyWithReaction` all route through `CallParticipantState._`, so two allocations plus a 10-element copy now happen on paths that have nothing to do with audio levels: viewport visibility, paused tracks, connection quality, pins, reactions. In a 100-participant call that is 100 unnecessary list copies per state write — in a PR whose purpose is reducing allocation pressure.

The comment's concern (a view writing through to a caller-held backing list) is real, but only for the public constructor. The four internal call sites always pass a list that is already sealed or freshly built. Skipping the copy when the incoming list is already one of ours would recover the cost.

packages/stream_video/lib/src/call/state/mixins/state_sfu_mixin.dart
```diff
@@ -128,22 +159,34 @@ mixin StateSfuMixin on StateNotifier<CallState>, StatePendingTracksMixin {
...
+            participant.userId,
+            participant.sessionId,
+          )];
+
+      // A participant the event does not mention, or who was silent and still
+      // is, keeps their existing instance. The event that takes them below the
+      // speaking threshold is still applied, so their `audioLevel` and
+      // `audioLevels` hold at that reading until they speak again.
+      if (levelInfo == null ||
+          (!levelInfo.isSpeaking && !participant.isSpeaking)) {
```
Brazol, line 181: **The hold-while-silent rule is applied on only one of the two paths that write audio levels.**

`sfuParticipantUpdated` still calls `copyWithUpdatedAudioLevels` unconditionally, so a silent participant's `audioLevel` and `audioLevels` **do** advance on that event, while this handler holds them. Worth making consistent, or documenting why the two differ.

Separately, on the behaviour change itself: nothing in `stream_video_flutter` reads `audioLevel` or `audioLevels`, so the only people affected are integrators drawing their own audio meter, whose meters will now stick at the last reading instead of decaying to zero. It is in the CHANGELOG and `isSpeaking` is the documented escape hatch, so this is a judgement call rather than a bug — but it will generate a support ticket.

packages/stream_video_flutter/lib/src/call_participants/call_participants.dart
renefloor: Agreed and fixed in 93c7f869.

All three sites now pass an `onError` that logs the error and stack and leaves the subscription intact, so the last known list stays on screen and later updates keep landing. The two in `StreamCallParticipants` collapsed into one `_subscribeToParticipants()` helper rather than repeating the handler.

One thing worth recording for anyone reading this later: reachability is narrow, since the default tier function cannot throw and this needs an integrator-supplied resolver. But it is not a one-shot when it does happen — `adaptive_throttle`'s `onData` returns after `addError` without arming a window, so the next upstream value re-enters and throws again. It was an uncaught error per participant event, not one.

Covered by `keeps the last list when participantsStream errors`, which fails on the pre-fix widget: `testWidgets` reports the zone error as a test failure, which is exactly the symptom.

packages/stream_video_flutter/lib/src/call_participants/call_participants.dart
renefloor: Agreed and fixed in 93c7f869, on both branches.

Your causal account is right — I checked it: pre-PR, `copyWithUpdatedAudioLevels` produced a new Equatable value on every audio tick, so the list emitted continuously and `recalculateParticipants` picked up the new comparator within ~100 ms. This PR stops those writes, so the mitigation is gone.

I extended it past the one clause you suggested. The controlled-`participants` branch had the same gap: a `sort` change there was stuck behind the `ListEquality` check, which compares the participants and knows nothing about the comparator. So `orderingChanged` is computed once and consulted on both branches.

On the inline-closure trade-off: I took it. The identity guard in `recalculateParticipants` absorbs it into no `setState`, so the cost is an O(n log n) sort on rebuild, and comparing callbacks in `didUpdateWidget` is ordinary Flutter practice. Noted in the code comment so the next person doesn't have to re-derive it.

Covered by `re-sorts when the sort comparator changes` and `re-filters when the filter changes`, both of which fail on the pre-fix widget.

packages/stream_video/lib/src/models/call_participant_state.dart
renefloor: Agreed and fixed in 93c7f869.

Two adjustments to the framing, one down and one up.

**Down:** the count tracks *changed* participants, not list length — `_updateParticipants` skips the ones it hands back untouched, so `copyWith` never runs for them. It reaches 100 on a connection-quality or inbound-state event that genuinely touches everyone, not on every state write.

**Up:** the hottest path was worse than the ones you listed. `copyWithUpdatedAudioLevels` built `[...audioLevels, audioLevel]`, then `removeRange`d it, and then `_sealLevels` copied *that* again into a view — four lists per speaking participant per audio event, in the handler this PR exists to optimize.

On the fix: I didn't take `is UnmodifiableListView` as the test, because it isn't safe. A caller can hand you an `UnmodifiableListView` over a list they still hold — precisely the write-through hole the seal was added to close, and something the existing test `cannot be mutated through an unmodifiable view of a live list` already pins down. So the marker is a private `_SealedLevels extends UnmodifiableListView<double>`; being private is the proof that the backing list came from inside the class and is unreachable from anywhere else.

Result: non-audio `copyWith` allocates nothing for levels, and the audio path builds the window with the front already dropped and seals by adoption — two lists where it used to be four.

Covered by `a copy that leaves audio alone shares the level list` (fails pre-fix), plus tests that a level update does *not* share history and that a sealed list is still unmodifiable after an unrelated `copyWith`. Both write-through tests you'd expect to break still pass.

packages/stream_video/lib/src/call/state/mixins/state_sfu_mixin.dart
renefloor: Agreed and fixed in 93c7f869.

You're right, and it was worse than the comment says: `sfuParticipantUpdated` didn't route through `_updateParticipants` at all. It mapped, always allocated a new instance for the matched participant, and always wrote state. So it was missing the no-op guard as well as the audio rule — and the PR description's claim that it "skip[s] the state write entirely when the event leaves everything unchanged" was simply wrong for that handler. I've corrected the description too.

It now goes through `_updateParticipants` with the same `!isSpeaking && !wasSpeaking` guard, so the two paths agree. The guard falls back to an Equatable compare to decide whether anything else in the event changed — that's one deep compare on a cold event, not on the audio tick path, so it buys the no-op skip cheaply.

On your second point: I kept the behaviour change as-is. It's in the changelog, `isSpeaking` is the documented escape hatch, and it matches Swift. But your framing is what convinced me the fix was necessary rather than optional — the "judgement call" defence only holds if the rule is actually uniform, which it wasn't.

Covered by four new tests: holding a silent participant's level, advancing it once they speak, still writing a changed field through, and leaving the other participants on their instances. The first fails on the pre-fix handler.

Brazol (approved):

